PR #81, opened 2023-01-20
feat: support async components using Suspense

stevenle:


stevenle:
Closing, I don't think we'll be able to support this natively until there's a way to handle this without two passes.

PR #130, opened 2023-04-16
feat: add rds package with <Text> and <Heading>

stevenle:
## Background

These two foundational typography components will be the first two components for the Root.js Design System (RDS) and will demonstrate how a project will generally depend on the RDS package and use it within their projects.

## Objective

The `<Text>` component is a server-rendered component that applies typography styles to an element. The `<Heading>` component is a lightweight wrapper around the <Text> component that specifically outputs `<h1>`, `<h2>`, etc. semantic tags.

The typography styles should be configurable through design tokens using a SCSS map to store configuration values, which may include type sizes (font-size, line-height, letter-spacing), weights (font-weight), transforms (uppercase), family (font-family), and responsive styles based on breakpoints.

These typography components should also be able to transform markdown to safe HTML and apply default styles for any child elements (e.g. `<ul>` and `<ol>` lists).

From a build standpoint, the RDS package will need to be distributed as a "source" package that will eventually be bundled and transformed through the project's root bundler. This is so that the components can read from the project's design tokens config directly and generate the necessary style variations.

## Build plan doc

https://docs.google.com/document/d/103O6aGvQ19Wflb8yfMTsukKlcHwc3vkHZkKDlJydwkw/edit#

### DIFF
--- a/packages/rds/src/components/Heading/Heading.tsx
+++ b/packages/rds/src/components/Heading/Heading.tsx
@@ -1,13 +1,32 @@
 import {ComponentChildren} from 'preact';
 
-import {Text, FontWeight} from '../Text/Text.js';
+import {Text, FontWeight, TextSize} from '../Text/Text.js';
 
 export type HeadingLevel = 1 | 2 | 3 | 4 | 5 | 6;
 export type HeadingComponent = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
 
+/**
+ * The heading size values are generated from a variable called `$text-sizes` in
+ * `tokens.scss`.
+ *
+ * Example:
+ *
+ * ```
+ * // tokens.scss
+ * $text-sizes: (
+ *   'title-lg': (
+ *     'font-size': 32px,
+ *     'line-height': 1.3,
+ *     'font-weight': 700,
+ *   ),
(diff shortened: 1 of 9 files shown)


packages/rds/src/components/Heading/Heading.tsx
```diff
@@ -0,0 +1,27 @@
+import {ComponentChildren} from 'preact';
+
+import {Text, FontWeight} from '../Text/Text.js';
+
+export type HeadingLevel = 1 | 2 | 3 | 4 | 5 | 6;
+export type HeadingComponent = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+
+export type HeadingProps = preact.JSX.HTMLAttributes & {
+  level?: HeadingLevel;
+  size?: string;
```
jeremydw, line 10: If this is just `string` it doesn't give any hints to the possible values. Maybe we create in `Text.tsx` something like the following. The cherry on top is if we can emit (or generate) the possible values somehow for stronger validation/checking.

```
/** 
 * Describe what the possible values are and how to use tokens.
 */
type Size = string;
```

jeremydw (approved):

jeremydw (approved):

PR #171, opened 2023-08-01
feat: add rds <Image> component

stevenle:
Features:

* Outputs just a `<picture>` tag, no additional element wrappers needed
* The styles for the component are wrapped in a nested cascade layer called `@layer rds.component`, default css styles for the image can be easily overridden simply by targeting the `img` selector
* If `preload={true}` is passed, automatically injects `<link rel="preload">` tags into `<head>`
* Automatically transforms images through an `ImageService`, which supports App Engine legacy Images API by default and can be configured to support other transform services via Context API
* User can request specific sizes of the image using the `sizes` prop, which accepts either a number (a single width for all breakpoints) or a breakpoint map, e.g. `sizes={{default: 1200, mobile: 500, tablet: 1000}}`

Note: instead of using a custom element to detect the image's actual container width, using pre-defined sizes for each breakpoint is preferred here so that the image size is deterministic when the HTML for the page is fetched. This allows the preload tag to work properly since image preloading would not be aware of the image's container size.


packages/rds/src/components/Image/Image.module.scss
```diff
@@ -0,0 +1,12 @@
+@layer rds.component {
+  .picture {
```
jeremydw, line 2: I've had a lot of success using the old "FifeImage" component but without any defined styles (letting the caller specify styles, and having the emitted `` behave like the default `` without any styling from the core component. Do you think we need the `Image` component to come with styles?

stevenle: The styles here are pretty minimal and would be the same reset style I generally use for `img` tags. I can remove these styles here in favor of having a base stylesheet that normalizes our `img` (and `svg`, `video`) styles.

jeremydw: Yeah; I think it'd be better to have the `<Image>` component not apply styles (at least by default). For example, if I wanted to have an inline image appear without taking up the full width of the container I'd need to overwrite these styles.

stevenle: Counter-point: in more cases than not, I'd expect the image to want to take the full-width of its container, and only in cases where I don't want it to would I override the style (to either `width: auto` or `width: 50%` or `max-width: 500px`). I usually think if a user uploads an image smaller than the container and it doesn't fill the whole width, it'd be considered a bug.

But again, I'm happy to defer that to a base stylesheet.

PR #172, opened 2023-08-01
feat: base.scss styles for rds

stevenle:
Build plan doc with previous discussion/comments:
https://docs.google.com/document/d/1tA5A04pH7asOHjdQZo_1t2g55fOlfNZWvL5p5eVdfHw/edit

The base.scss file is meant to act similarly to a "reset stylesheet" to provide consistency across all components. For the most part, only styles that are useful for ALL projects should be included in this sheet. Advanced options (like smooth-scroll and prefers-reduced-motion animation timing resets) can be disabled via SCSS configuration variables.

### DIFF
--- a/packages/rds/src/styles/default.scss
+++ b/packages/rds/src/styles/default.scss
@@ -1,0 +1,99 @@
+/**
+ * Default CSS styles. This file is similar to a "reset" stylesheet and
+ * represents base styles for consistency across all RDS components.
+ *
+ * This stylesheet can be imported in two different ways:
+ *
+ * 1) Import within `BaseLayout.tsx`:
+ *
+ * ```
+ * import '@blinkk/rds/styles/default.scss';
+ * ```
+ *
+ * Note: Cannot configure the styles using this method. See (2) if you need to
+ * configure the CSS output.
+ *
+ * 2) Import within a SCSS file, e.g. `global.scss`:
+ *
+ * ```
+ * @use '@blinkk/rds/styles/default.scss';
+ * ```
+ *
+ * To change any configurations, use the `@use ... with` syntax, e.g.:
+ *
+ * ```
+ * @use '@blinkk/rds/styles/default.scss' with (
+ *   $disable-prefers-reduced-motion: true;
+ * );
+ * ```
+ */
+
+$disable-smooth-scroll: false !default;
+$disable-prefers-reduced-motion: false !default;
+
+@layer rds.base {
+  //  Use a more intuitive box-sizing strategy.
+  *,
+  *::before,
+  *::after {
+    box-sizing: border-box;
+  }
+
+  // Reset margins.
+  body {
+    margin: 0;
+  }
+
+  // The default `display: inline` property of img and video causes extra
+  // vertical space within a parent container. `display: block` is generally
+  // more intuitive. Also set a max-width and auto height based on the
+  // width and height attributes of the element.
+  img,
+  svg,
+  video {
+    display: block;
+    max-width: 100%;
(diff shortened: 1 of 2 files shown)


packages/rds/src/styles/default.scss
```diff
@@ -0,0 +1,99 @@
...
+ * @use '@blinkk/rds/styles/default.scss' with (
+ *   $disable-prefers-reduced-motion: true;
+ * );
+ * ```
+ */
+
+$disable-smooth-scroll: false !default;
+$disable-prefers-reduced-motion: false !default;
+
+@layer rds.base {
```
uxder, line 34: @layer is relatively new to the team.

If we want to override any of the rds.base layer, I think we would also need to wrap it in the project level layer otherwise it won't win the specificity battle. 

Maybe we should add an example of that?

```
@use '@blinkk/rds/styles/default.scss';
@layer my-project {
   img {
     max-width: none;
   }
}
```

stevenle: When no cascade layer is set, it wins the specificity battle over all other cascade layers. Doing this would be enough to override the rds.base layer:

```
img {
  max-width: auto;
}
```

uxder: Ah gotcha.  I'm still picking up my knowledge on layers.  Thanks for confirming :)

uxder (approved):

packages/rds/src/styles/default.scss
```diff
@@ -0,0 +1,99 @@
...
+
+  // Set smooth scrolling by default. The `:focus-within` helps with cmd+f
+  // functionality.
+  // https://css-tricks.com/fixing-smooth-scrolling-with-find-on-page/
+  @if not($disable-smooth-scroll) {
+    html:focus-within {
+      scroll-behavior: smooth;
+    }
+    html {
+      --scroll-padding-top: var(--header-height, 0px);
```
jeremydw, line 75: We don't need to handle this right now but I'm wondering if we want to adopt a prefixing strategy for our CSS vars, e.g. `--rds-scroll-padding-top`.

I look at the styles for https://github.com/ in incognito to get the marketing page (GitHub's been a long time user of CSS vars and tokens), and they do NOT use such prefixes. Stripe doesn't either. So maybe we don't -- just wanted to bubble it up for discussion.

stevenle: I gave it some thought, I think generally when I see a prefix like `--rds-` I'd see it as an encapsulated configuration variable that's either used internally by the component or is exposed by the component as an API. So like, if for example I have an `<rds-button>` component, maybe the rds-button would have a CSS variable like `--rds-button-icon-size` to provide configuration options to the user.

For general-purpose variables that can be used throughout the site, I think I prefer non-rds-prefixed variables. Feels a little cleaner and doesn't feel like I'm reaching in and modifying something that I shouldn't be modifying.

I took a look at your github.com example and I really liked how clean some of their base CSS variables are, e.g.

```
--h1-size: 2rem;
--breakpoint-small: 34rem;
```

That said, they're not super consistent and in some cases they do prefix with the word `base` (I assume "base" is the name of their design system.), e.g.:

```
--base-text-weight-light: 300;
```

For now, my recommendation is to prefix based on functionality, and reserve the `--rds-` prefix for configuring `<rds-*>` custom elements.

stevenle:
planning to submit this tomorrow, lmk if anyone has any add'l feedback.

stevenle:
FYI I'm going to plan to merge this in by EOD, let me know if any add'l comments!

PR #176, opened 2023-08-03
feat: add rds breakpoints.scss mixins

stevenle:
The RDS breakpoints mixins use a device naming convention with the following:

* sm (mobile)
* md (tablet)
* lg (laptop)
* xl (desktop)
* xxl (site-max)

The widths of these breakpoints can be configured in `tokens.scss`, but the names need to be fixed since SCSS doesn't have a way to dynamically generate mixins (unless we want to switch to calling our mixins like `@include breakpoints.at('mobile') { /* styles */ }`, which from past discussions sounds undesirable).

One benefit of using a fixed set of breakpoint names is that other RDS components can be written in a consistent way without having to assume any number of breakpoint possibilities.

### DIFF
--- a/packages/rds/src/styles/breakpoints.scss
+++ b/packages/rds/src/styles/breakpoints.scss
@@ -7,7 +7,7 @@
  * @use 'breakpoints';
  * .foo {
  *   color: blue;
- *   @include breakpoints.mobile() {
+ *   @include breakpoints.sm() {
  *     color: red;
  *   }
  * }
@@ -18,7 +18,7 @@
  * ```
  * .foo {
  *   color: blue;
- *   @include breakpoints.tablet-lt() {
+ *   @include breakpoints.md-lt() {
  *     color: red;
  *   }
  * }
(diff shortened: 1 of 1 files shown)


jeremydw:
Time to bite the bullet and adopt sm, md, lg, xl as the names for the names instead? (I waffle on this personally as I've enjoyed the other breakpoint names for years; although on the one project where I use the size-based breakpoint names I've found the transition easier than expected.)

stevenle:
> Time to bite the bullet and adopt sm, md, lg, xl as the names for the names instead? (I waffle on this personally as I've enjoyed the other breakpoint names for years; although on the one project where I use the size-based breakpoint names I've found the transition easier than expected.)

For the record, I hate it, but it's probably the right choice. I checked tailwind, bootstrap, and a few others, and pretty much everyone else uses those names for breakpoints.

stevenle:
Another thing we might want to consider is how to name the `-gt` and `-lt` mixins. With the device naming convention (`laptop-gt()`), I always read it as "laptop and greater", but someone new to the size mixins might find the acronyms to be gibberish (`xl-lt()`). I don't have a better suggestion. boostrap uses `-up()` and `-down()`, but I'm not sure if those are any better.

PR #339, opened 2024-05-31
feat: add StringParamsProvider for injecting param values to strings

stevenle:
For sites that use parameterized values throughout the site, a developer can configure a context variable that can automatically be used in any strings that use the `useTranslations()` hook. For example:

```jsx
export default function Page() {
  const stringParams = {name: 'Alice'};
  return (
    <StringParamsProvider value={stringParams}>
      <SayHello />
    </StringParamsProvider>
  );
}

function SayHello() {
  const t = useTranslations();
  return <h1>{t('Hello, {name}!')}</h1>;
}

// =>

<h1>Hello, Alice!</h1>
```


stevenle:
@jeremydw The main use case for this is to manage pricing variables separately from the normal translations flow. Chose "GlobalParams" here as a generic name, but if you have other name suggestions let me know. 

The developer would need to fetch the pricing info from somewhere and then pass it to the `<GlobalParamsProvider>`, which is then auto-injected as param values whenever `useTranslations()` is used.

stevenle:
After some discussion with all the GPTs, decided to rename to `StringParamsProvider`. Merging this in now.

PR #343, opened 2024-06-04
feat: add publishing lock

### LINKED ISSUE #341: Disable publishing
**Description**

For sensitive launches, there should be a dropdown menu option to "disable publishing" with options for having it disabled either indefinitely or until a certain date/time.

When a page is publish-disabled, the publish button should show a lock icon with some indication of who locked it and the reason why it's locked.

stevenle:
fixes #341 

screens:

Screenshot 2024-06-03 at 11 14 58 PM

Screenshot 2024-06-03 at 11 15 18 PM

Screenshot 2024-06-03 at 11 15 25 PM

Screenshot 2024-06-03 at 11 15 30 PM

Screenshot 2024-06-03 at 11 21 32 PM

### DIFF
--- a/.changeset/fresh-countries-brush.md
+++ b/.changeset/fresh-countries-brush.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: add publishing lock
--- a/packages/root-cms/core/client.ts
+++ b/packages/root-cms/core/client.ts
@@ -7,7 +7,6 @@ import {
   Query,
   Timestamp,
   WriteBatch,
-  getFirestore,
 } from 'firebase-admin/firestore';
 import {CMSPlugin} from './plugin.js';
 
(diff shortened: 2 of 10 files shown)


jeremydw (approved):
Feature and UI strings LGTM - thank you!

One additional thing to consider is whether this should be built into a future "checks" system (like a test suite) that gets run during document publishing.
The publishing flow modal can show all the failed (or successful checks), e.g.

- Locked for publishing
- Blocked terms
- Site-specific custom checks
- Approval status
- etc.

Not sure if this alters how this is implemented but just wanted to bubble up that future concept.

stevenle:
Yeah, I've been thinking about how best to incorporate an approval flow into the UI in a way that's intuitive. My thinking right now is that if an approval config is enabled, we would force all publishing to go through the "Release" process and the "Publish" button on the document page would be changed to "Add to release" (or "Go to release"). The release can contain 1 or more doc, a description of the release, and snapshots of before/after for each doc in the release. This would be similar to a GitHub PR - we can maybe even enable comments by allowing people to comment within the snapshot.

Within this release review system, we can enable "presubmit checks" like you're suggesting (e.g. checking for blocked terms).

PR #466, opened 2025-02-24
fix: output correct types for oneOf field

jeremydw:


stevenle:
add a changeset:

* run `pnpm changeset`
* scroll down to `@blinkk/root-cms` and press spacebar to select
* when it asks "which package major/minor bump" press enter twice to ignore (this will mark it a "patch" bump)
* enter in the summary (the same as the PR description)

then commit the file and push it to this PR (a regular commit is fine, i'll squash the commits in the PR anyway)

jeremydw:
Done

jeremydw:
should be updated now

PR #467, opened 2025-02-24
feat: add hotkey support to array items in doc editor

jeremydw:
screencast:
https://github.com/user-attachments/assets/026412b9-6ef3-4a21-abd1-367a99f3ac0a

- the focus ring is only visible once you use the hotkeys
- otherwise, the only ui change is the last-moved field gets underlined (this helps disambiguate between two fields with the same preview)

### DIFF
--- a/packages/root-cms/ui/components/DocEditor/DocEditor.tsx
+++ b/packages/root-cms/ui/components/DocEditor/DocEditor.tsx
@@ -507,6 +507,8 @@ DocEditor.ObjectFieldDrawer = (props: FieldProps) => {
 interface ArrayFieldValue {
   [key: string]: any;
   _array: string[];
+  /** Tracks the last-moved array item. */
+  _moved?: string;
   _new?: string[];
 }
 
@@ -681,6 +683,7 @@ function arrayReducer(state: ArrayFieldValue, action: ArrayAction) {
       return {
         ...data,
         _array: order,
+        _moved: order[action.index - 1],
       };
     }
     case 'moveDown': {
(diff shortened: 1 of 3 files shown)


packages/root-cms/ui/components/DocEditor/DocEditor.tsx
```diff
@@ -840,7 +860,30 @@ DocEditor.ArrayField = (props: FieldProps) => {
             key={key}
             open={newlyAdded.includes(key) || itemInDeeplink(key)}
           >
-            <summary className="DocEditor__ArrayField__item__header">
+            <summary
+              className="DocEditor__ArrayField__item__header"
+              tabIndex={0}
+              id={`summary-for-${props.deepKey}.${order[i]}`}
+              onKeyDown={(e: KeyboardEvent) => {
```
stevenle, line 867: since there's a lot of logic here, let's move this callback into a function so that there's not so much inlining for this prop value.

jeremydw:
should be ready for another look now

packages/root-cms/ui/components/DocEditor/DocEditor.tsx
```diff
@@ -828,6 +848,26 @@ DocEditor.ArrayField = (props: FieldProps) => {
     );
   }
 
+  /** Handler for using the arrow keys when the array item's header is focused.  */
+  function handleKeyDown(e: KeyboardEvent, key: string) {
```
stevenle, line 852: `key` is pretty ambiguous here, maybe change to `arrayKey` instead.

Issue #437, opened 2024-10-18
Copy/paste hotkeys


stevenle:
**Description**

Add support for hotkeys for copy/pasting certain field editor items, e.g. array items and images.

PR #468, opened 2025-02-24
feat: add preliminary virtual clipboard to arrays

jeremydw:
- enables quick copying and pasting before/after via the menu

screencast:

https://github.com/user-attachments/assets/edad294c-b5f0-4ec6-bace-4e642718e970

#437


stevenle:
oh wow very nice. thanks for the feature!

PR #474, opened 2025-03-04
fix: subtle race condition in createProdServer

jeremydw:


packages/root/src/cli/start.ts
```diff
@@ -69,7 +69,7 @@ export async function createProdServer(options: {
   server.use(sessionMiddleware());
 
   const plugins = rootConfig.plugins || [];
-  configureServerPlugins(
+  await configureServerPlugins(
```
stevenle, line 72: can you add the same to `cli/preview.ts`?

jeremydw: Done

PR #477, opened 2025-03-25
fix: remove sourcemap from cms static files

jeremydw:
- also modify cachebusting approach in non-local envs
- this is meant to be a surgical change to improve cms load times


packages/root-cms/core/app.tsx
```diff
@@ -277,7 +278,13 @@ function getRefererOrigin(req: Request): string {
...
 
-function cachebust(url: string) {
-  const ts = Math.floor(new Date().getTime() / 1000);
-  return `${url}?c=${ts}`;
+/** Modify the given URL to bust the cache. */
+function cachebust(req: Request, url: string) {
+  // In local dev, use the the timestamp to cachebust.
+  // In non-local mode, use the package version.
+  const value =
+    req.hostname === 'localhost'
```
stevenle, line 286: can you confirm if hostname would ever include the port?

packages/root-cms/core/app.tsx
```diff
@@ -3,6 +3,7 @@ import path from 'node:path';
 import {fileURLToPath} from 'node:url';
 import {Request, Response, RootConfig} from '@blinkk/root';
 import {render as renderToString} from 'preact-render-to-string';
+import packageJson from '../package.json' assert {type: 'json'};
```
stevenle, line 6: i'm a little uncertain about how this import works when compiled, will need to test/verify on a non-crucial project (e.g. on the rootjs.dev site).

packages/root-cms/core/app.tsx
jeremydw: Yeah; I tried this and it worked locally, so it should be good.

packages/root-cms/core/app.tsx
jeremydw: I ran the project locally and used it against the docs package and it seemed to work. I think the local build (via `pnpm run build` at the project root) bundles things the same way as in prod?

stevenle (approved):

stevenle:
last thing - add a changeset

Issue #484, opened 2025-03-31
Add GIF support for images


jeremydw:
We have a project where we need to upload GIFs.

There are a few places where GIF needs to be added to the list of supported types: 
https://github.com/search?q=repo%3Ablinkk%2Frootjs%20png&type=code

Additionally we need to figure out how to make GIFs compatible with the Google Image Service (they _are_ compatible as source files (https://cloud.google.com/appengine/docs/standard/services/images?tab=python#image-formats).

PR #485, opened 2025-03-31
feat: add gif support #484

jeremydw:


packages/root-cms/ui/components/DocEditor/fields/ImageField.tsx
```diff
@@ -8,13 +8,24 @@ import {joinClassNames} from '../../../utils/classes.js';
 import {uploadFileToGCS} from '../../../utils/gcs.js';
 import {FieldProps} from './FieldProps.js';
 
-export const IMAGE_MIMETYPES = [
-  'image/png',
+/** Mimetypes accepted by the image input field. */
+const IMAGE_MIMETYPES = [
```
stevenle, line 12: Only keep `images/gif` here, for all the others we don't want to encourage use of them by default. Also image/jpg is not a valid mimetype so please remove.

jeremydw: Done

packages/root-cms/ui/components/DocEditor/fields/ImageField.tsx
```diff
@@ -8,13 +8,20 @@ import {joinClassNames} from '../../../utils/classes.js';
...
+  'image/gif',
   'image/jpeg',
+  'image/png',
   'image/svg+xml',
   'image/webp',
 ];
 
+/** Builds the `accept` attribute for the `input` field.*/
+export function buildAcceptAttribute(exts?: string[]) {
```
stevenle, line 21: Remove the "export" since I think this is only used in this file.

jeremydw: Done

PR #497, opened 2025-04-15
fix: improve arb download feature

jeremydw:
- add preview feature for auditing strings before downloading
- download arb as attachment file instead of blob

cc: @adelaidemt


packages/root-cms/ui/pages/TranslationsArbPage/TranslationsArbPage.tsx
```diff
@@ -148,3 +207,31 @@ function parseDocIds(docIdsText: string): string[] {
...
+  )}t${pad(now.getHours())}${pad(now.getMinutes())}`;
+  const projectName =
+    window.__ROOT_CTX.rootConfig.projectName ||
+    window.__ROOT_CTX.rootConfig.projectId;
+  const parts = [slugify(projectName)];
+  if (docIds.length === 1) {
+    parts.push(slugify(docIds[0]));
+  }
+  parts.push(timestamp);
+  return `${parts.join('_')}.json.arb`;
```
stevenle, line 229: is .json.arb a standard thing? if not, i'd prefer to keep it clean as just .arb.

PR #513, opened 2025-04-28
chore: make select fields searchable

jeremydw:
this helps with long lists of values; the user can hit the first letter of the template to jump to it

### DIFF
--- a/packages/root-cms/ui/components/DocEditor/DocEditor.tsx
+++ b/packages/root-cms/ui/components/DocEditor/DocEditor.tsx
@@ -1169,6 +1169,7 @@ DocEditor.OneOfField = (props: FieldProps) => {
           onChange={(e: string) => onTypeChange(e || '')}
           size="xs"
           radius={0}
+          searchable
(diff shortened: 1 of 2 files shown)


packages/root-cms/ui/components/DocEditor/DocEditor.tsx
```diff
@@ -1169,6 +1169,7 @@ DocEditor.OneOfField = (props: FieldProps) => {
           onChange={(e: string) => onTypeChange(e || '')}
           size="xs"
           radius={0}
+          searchable
           // Due to issues with preact/compat, use a div for the dropdown el.
           dropdownComponent="div"
```
stevenle, line 1174: were you able to test this change? reason i ask is because the `dropdownComponent="div"` was added because of some incompatibility issues and so i wanna make sure this works with the div implementation.

jeremydw: Yeah; I gave it a try on the `docs` package and it LGTM. It's a little quirky in that it supports substring matches but that actually makes it easier to locate the desired value

PR #552, opened 2025-05-21
feat: Add pre/post string transformations for useTranslations

stevenle:
This commit introduces a mechanism to transform strings before and after they are processed by the `useTranslations()` hook. This allows for greater flexibility in handling translations, such as overriding strings for specific locales or reformatting parameters.

Key changes:

- **Transformation Function Types**: Defined `PreTranslationFunc` and `PostTranslationFunc` types in `packages/root/src/core/types.ts`.
- **TransformationContext**: Created `TransformationContext` and a `useTransformationContext` hook (`packages/root/src/core/hooks/useTransformationContext.ts`) to provide transformation functions. It defaults to identity functions if no provider is used.
- **`useTranslations` Update**: Modified the `useTranslations` hook (`packages/root/src/core/hooks/useTranslations.ts`) to consume `TransformationContext`. It now applies `preTranslation` before translation lookup and `postTranslation` after parameter substitution. The current locale is passed to both transformation functions.
- **`TransformationProvider`**: Added a `TransformationProvider` component (`packages/root/src/core/components/TransformationProvider.tsx`) that allows you to supply your custom `preTranslation` and `postTranslation` functions.
- **Documentation**: Updated TSDoc for all new and modified components/hooks, including examples for pre- and post-transformation use cases (e.g., locale-specific overrides, parameter reformatting, `&nbsp;` injection).
- **Unit Tests**: Added a comprehensive suite of unit tests in `packages/root/test/translations.test.ts` to cover various scenarios, including transformations, locale handling, and default behavior.

This addresses the need for middleware-like string transformations within the localization process, enabling developers to:
- Override source strings for different locales (e.g., EN vs. ROW).
- Reformat parameters (e.g., `foo[footnote:bar]` to `foo{footnote:bar}` or `foo<sup>bar</sup>`).
- Inject HTML entities like `&nbsp;` for product names.

### DIFF
--- a/packages/root/src/core/components/TransformationProvider.tsx
+++ b/packages/root/src/core/components/TransformationProvider.tsx
@@ -1,0 +1,135 @@
+import {FunctionComponent, ReactNode} from 'preact';
+import {TransformationContext} from '../hooks/useTransformationContext.js';
+import {PreTranslationFunc, PostTranslationFunc} from '../types.js';
+
+/**
+ * Props for the `TransformationProvider` component.
+ */
+export interface TransformationProviderProps {
+  /**
+   * An optional function to be called before a string is translated.
+   * This function receives the original string and the current locale.
+   * It can be used to modify the string before it's looked up in
+   * translation files, for example, to normalize text or provide
+   * locale-specific string variations.
+   * If not provided, the original string is used directly.
+   */
+  preTranslation?: PreTranslationFunc;
+  /**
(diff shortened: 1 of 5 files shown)


stevenle:
notes:

the code produced by codex (#551) was slightly closer to the implementation i had in mind, although the tests here seem more comprehensive. abandoning this pr in favor of the latter, will migrate the tests later at some point when i have a chance.

PR #582, opened 2025-06-16
Add sirvOptions config

jeremydw:
## Summary
- allow specifying `sirvOptions` in the server config
- merge the custom Sirv options with Root.js defaults for dev, start and preview commands

## Testing
- `pnpm test` *(fails: Request was cancelled - network access blocked)*

------
https://chatgpt.com/codex/tasks/task_e_68509d2094a48322ae767457ada2443c


stevenle:
code looks fine. i'm debating whether we want to rename it to something else since "sirv" is an implementation detail and in the future we may want to swap it out for something else. 

in any case we can leave it as is for now, lmk when you've tested this and i can merge.

PR #621, opened 2025-07-30
Add ability to expand/collapse preview and improve iframe loading

jeremydw:
Threw away most of what Copilot did in #610.

This PR adds the ability to expand/collapse the preview:

- Adds expand/collapse button next to the "Save" button
- Makes no other UI changes (keep JSON inside the three dots menu)
- Tidy up Edit JSON dialog

This PR also defers the iframe loading until after the fields have rendered, in order to reduce resource consumption that occurs when both rendering the fields _and_ loading the iframe. This is useful for pages with intensive fields and intensive previews.

Default:

image

Hidden:

image

### DIFF
--- a/packages/root-cms/ui/components/DocEditor/DocEditor.tsx
+++ b/packages/root-cms/ui/components/DocEditor/DocEditor.tsx
@@ -58,6 +58,10 @@ import {extractField} from '../../utils/extract.js';
 import {getDefaultFieldValue} from '../../utils/fields.js';
 import {flattenNestedKeys} from '../../utils/objects.js';
 import {autokey} from '../../utils/rand.js';
+import {
+  cancelIdleCallbackPolyfill,
+  requestIdleCallbackPolyfill,
+} from '../../utils/request-idle-callback.js';
 import {getPlaceholderKeys, strFormat} from '../../utils/str-format.js';
 import {testFieldEmpty} from '../../utils/test-field-empty.js';
 import {formatDateTime} from '../../utils/time.js';
@@ -151,12 +155,12 @@ export function DocEditor(props: DocEditorProps) {
   // Notify parent when fields are rendered (for deferred iframe loading)
   useEffect(() => {
     if (!loading && props.onFieldsRendered) {
-      // Use a small delay to ensure fields are actually rendered in DOM
-      const timer = setTimeout(() => {
-        props.onFieldsRendered();
-      }, 100);
-      return () => clearTimeout(timer);
+      const timer = requestIdleCallbackPolyfill(() => {
+        props.onFieldsRendered && props.onFieldsRendered();
+      });
+      return () => cancelIdleCallbackPolyfill(timer as number);
     }
+    return () => {};
   }, [loading, props.onFieldsRendered]);
 
   return (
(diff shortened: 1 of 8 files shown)


packages/root-cms/ui/components/DocEditor/DocEditor.tsx
```diff
@@ -144,6 +152,17 @@ export function DocEditor(props: DocEditorProps) {
     }
   }, [loading]);
 
+  // Notify parent when fields are rendered (for deferred iframe loading)
+  useEffect(() => {
+    if (!loading && props.onFieldsRendered) {
+      const timer = requestIdleCallbackPolyfill(() => {
+        props.onFieldsRendered && props.onFieldsRendered();
```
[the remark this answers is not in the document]
stevenle: Note: In addition to it being redundant here, I'm also not a fan of the pattern `fn && fn()` because of readability, so it'd be better to use a normal `if (fn) { fn() }` pattern instead.

packages/root-cms/ui/components/DocEditor/DocEditor.tsx
```diff
@@ -144,6 +152,17 @@ export function DocEditor(props: DocEditorProps) {
...
+      });
+      return () => cancelIdleCallbackPolyfill(timer);
+    }
+    return () => {};
```
stevenle, line 163: I don't think this return statement is necessary.

packages/root-cms/ui/pages/DocumentPage/DocumentPage.tsx
```diff
@@ -23,23 +25,62 @@ import {Layout} from '../../layout/Layout.js';
...
 interface DocumentPageProps {
   collection: string;
   slug: string;
 }
 
+/** Builds the preview URL for a document. */
+function getPreviewUrl(
+  collectionId: string,
+  slug: string,
+  selectedLocale = ''
```
stevenle, line 39:
```
options?: {locale?: string}
```

packages/root-cms/ui/utils/request-idle-callback.ts
```diff
@@ -0,0 +1,21 @@
...
+              return Math.max(0, 50 - (Date.now() - start));
+            },
+          });
+        }, 1);
+      };
+
+export const cancelIdleCallbackPolyfill =
+  typeof window !== 'undefined' && window.cancelIdleCallback
+    ? window.cancelIdleCallback
+    : function (id: number | NodeJS.Timeout) {
```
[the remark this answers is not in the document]
stevenle: NodeJS.Timeout is a server-side type, avoid using it in client-side code.

packages/root-cms/ui/pages/DocumentPage/DocumentPage.tsx
```diff
@@ -380,7 +490,11 @@ DocumentPage.Preview = (props: PreviewProps) => {
           </div>
         )}
         <div className="DocumentPage__main__previewFrame__iframeWrap">
-          <iframe ref={iframeRef} src={previewUrl} />
+          <iframe
+            ref={iframeRef}
+            src={props.allowIframePreview ? previewUrl : undefined}
```
stevenle, line 495: I dont' think the iframe tag should be output at all if the "allowIframePreview" is false

PR #624, opened 2025-07-31
feat: add drag and drop for list items

### LINKED ISSUE #619: Add ability to drag and drop array items in the CMS UI
**Description**

Let's add the ability to drag and drop array items in the CMS.

Here are the criteria it should follow:

1. Ensure the drag area is extremely explicit, for example, the array items should have a "handle" on the left, next to the disclosure triangle. Only the handle should be draggable.

Image

2. When dragging an array item, the "dragover" state should display just a black line clearly indicating where the array item will be placed upon release. Don't show the entire array item relocated during the drag period, to avoid layout shift.

3. Once releasing the drag, the array item should be relocated. Don't expand/collapse anything - that should all remain in the state it was pre-drag.

For now, we should only allow dragging and dropping within the same array parent (i.e. don't allow dragging to a different parent).

jeremydw:
Fixes #619 

- Used the same approach as https://ui.mantine.dev/category/dnd/

https://github.com/user-attachments/assets/dca82cd1-5564-443f-83bb-da814db717fb

### DIFF
--- a/packages/root-cms/package.json
+++ b/packages/root-cms/package.json
@@ -67,6 +67,7 @@
     "@genkit-ai/ai": "0.5.2",
     "@genkit-ai/core": "0.5.2",
     "@genkit-ai/vertexai": "0.5.2",
+    "@hello-pangea/dnd": "^18.0.1",
     "body-parser": "1.20.2",
     "commander": "11.0.0",
     "csv-parse": "5.5.2",
--- a/packages/root-cms/ui/components/DocEditor/DocEditor.css
+++ b/packages/root-cms/ui/components/DocEditor/DocEditor.css
@@ -184,10 +184,36 @@
   padding-left: 48px;
 }
 
-.DocEditor__ArrayField__items {
+.DocEditor__ArrayField__item__wrapper {
   display: flex;
-  flex-direction: column;
-  gap: 8px;
+  flex-direction: row;
+  gap: 2px;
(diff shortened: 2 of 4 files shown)


stevenle:
some small comments

- use `-E` to install the dependency (ensure `pnpm-lock.yaml` gets updated)
- add a changeset entry
- for the ux, i think having a box-shadow without a border always looks a little off to me, i think maybe i would try a simple grey border (i can tweak this later after submission)

jeremydw:
All should be updated now!

image

stevenle:
Thanks, I think we're close. There are a few small UX polish I'd like to fix before cutting a release for this, I can probably fix these in a separate PR just to get this merged.

After "dropping" an item I see a big blue outline which I think we can disable/hide:

Screenshot 2025-07-31 at 10 11 14 AM

Related to this I think we also have a subtle UI hint for focus visibility with text underline, which I think could be replaced with either bolded text or a subtle bg color.

The "drag" styles could also use a subtle CSS transition.

jeremydw:
@stevenle I tweaked it a bit to use a background color to signal focus instead of text style, outline, etc. I think it feels good but LMK your thoughts. Note that this style gets applied whenever you interact with the field (i.e. opening it, moving it with the keyboard, dragging it, or using the arrows). LMK if you think we should tweak that too.

PR #626, opened 2025-07-31
make array headers sticky when scrolling

jeremydw:
To provide the user with additional context about which field is open, make the array item headers sticky.

- Use an intersection observer since there is no `:stuck` selector
- Only the first level of array items get the sticky behavior and associated styles
- Suppress the arrows when the array item header is stuck

https://github.com/user-attachments/assets/8fb2f483-4856-402c-a6ad-c0bfa75ee041

### DIFF
--- a/packages/root-cms/ui/components/DocEditor/DocEditor.tsx
+++ b/packages/root-cms/ui/components/DocEditor/DocEditor.tsx
@@ -983,6 +983,34 @@ DocEditor.ArrayField = (props: FieldProps) => {
     }
   }
 
+  function createObserverForStuckHeader(summaryEl: HTMLElement) {
+    // Resolve the "stuck" height as the top bar, the sidebar header, and the height of the element itself.
+    const topBarHeight =
+      document.querySelector<HTMLElement>('.Layout__top')?.offsetHeight || 48;
+    const sideHeaderHeight =
+      getComputedStyle(summaryEl)
+        .getPropertyValue('--top-bar-height')
+        .trim()
+        .replace('px', '') || '36';
+    const elementHeight = summaryEl.offsetHeight;
+    const totalHeight =
+      parseFloat(sideHeaderHeight) + topBarHeight + elementHeight;
+    const observer = new IntersectionObserver(
+      ([entry]) => {
+        summaryEl.classList.toggle(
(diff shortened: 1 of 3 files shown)


packages/root-cms/ui/components/DocEditor/DocEditor.tsx
```diff
@@ -991,13 +1019,30 @@ DocEditor.ArrayField = (props: FieldProps) => {
         )}
         {order.map((key: string, i: number) => {
           const previewImage = arrayPreviewImage(field, value[key]);
+          const summaryRef = useRef<HTMLElement>(null);
+
+          useEffect(() => {
```
stevenle, line 1024: this pattern doesn't quite feel react-y, i wonder if we should create a custom component for the <details> element here so that this hook is isolated to the component.

i think a more common pattern is to create a hook instead of `createObserverForStuckHeader` and place the useEffect hook within the custom hook.

packages/root-cms/ui/hooks/useStuckObserver.ts
```diff
@@ -0,0 +1,35 @@
...
+        threshold: 0,
+        rootMargin: `-${height + elementHeight + 2}px 0px 0px 0px`,
+      }
+    );
+    observer.observe(ref.current);
+    return () => {
+      observer.disconnect();
+    };
+  }, []);
+  return {ref, isIntersecting};
```
stevenle, line 34: `isIntersecting` feels more of an internal implementation detail, i think it'd be better to use a more appropriate name like `isStuck`

packages/root-cms/ui/hooks/useStuckObserver.ts
```diff
@@ -0,0 +1,35 @@
+import {useEffect, useRef, useState} from 'preact/hooks';
+
+interface StuckObserverOptions {
+  offsetHeight: () => number;
+}
+
+/**
+ * Provides an observer that detects when an element is "stuck" at the top of the viewport.
+ */
+export function useStuckObserver(options: StuckObserverOptions) {
```
stevenle, line 10: Since this is for observing `position: sticky` elements I think I would call this `useStickyObserver()`

PR #628, opened 2025-07-31
Polish deeplink behavior

jeremydw:
- Enable using deeplinks without fully reloading the page
- Make a subtle fix for scrolling _to_ the deeplink
- Change the scroll behavior to "immediate" to remove the delay before focusing the field

@stevenle Mind taking a look? Let me know if you'd like me to _restore_ the smooth scrolling or if you want to leave it as "immediate".

https://github.com/user-attachments/assets/8a08dde7-76dc-4a5b-970b-9eadb92d8963

### DIFF
--- a/.changeset/slimy-snails-change.md
+++ b/.changeset/slimy-snails-change.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+fix: polish deeplink behavior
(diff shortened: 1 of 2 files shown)


stevenle:
just tested, this looks good, thanks for the fix. can you change one small thing before i merge?

PR #634, opened 2025-08-04
refactor: use new file upload field

jeremydw:


jeremydw:
- reduced the padding on the alt text box
- added a dedicated "upload file" action button next to the current icons, but i _also_ added the ability to double click the preview to trigger an upload. i didn't want to use the single click because then you won't be able to focus the field (and thus can't paste over it)
- confirmed mp4 preview works!

image

Here's a quick CSS edit of what the empty field looks like centered:

image

I'm going back and forth on this so LMK whichever you prefer and we can go with it. 

The default looks like this:

image

The reason I ultimately left-aligned it instead of center aligned it is because the other controls (add button, etc) are all left-aligned.

Also, I actually think we should _not_ make the aspect ratio of the empty field taller because for optional fields it might just take up too much space when needing to leave the field blank. But LMK if you feel strongly on this and we can change!

stevenle:
cool this looks good. i actually like the centered version a lot, let's go with that for now. tbh i didn't look too closely at the code but i can tweak things later so i can submit this if the centered version is already in this PR.

PR #639, opened 2025-08-05
feat: add experiment for enabling lexical editor

stevenle:


packages/root-cms/ui/components/RichTextEditor/lexical/LexicalEditor.tsx
```diff
@@ -0,0 +1,184 @@
...
+import {ShortcutsPlugin} from './plugins/ShortcutsPlugin.js';
+import {ToolbarPlugin} from './plugins/ToolbarPlugin.js';
+
+const INITIAL_CONFIG: InitialConfigType = {
+  namespace: 'RootCMS',
+  theme: LexicalTheme,
+  nodes: [
+    AutoLinkNode,
+    HeadingNode,
+    // ImageNode,
```
jeremydw, line 40: I've seen a few projects that use the EditorJs image embed; are they incompatible with Lexical at this time?

packages/root-cms/ui/components/RichTextEditor/lexical/LexicalEditor.tsx
```diff
@@ -0,0 +1,184 @@
...
+  const [floatingAnchorElem, setFloatingAnchorElem] =
+    useState<HTMLElement | null>(null);
+
+  // The "onRef" var is used as a callback, so it's typed to `any` here to avoid
+  // type warnings.
+  const onRef: any = (el: HTMLDivElement) => {
+    if (el) {
+      setFloatingAnchorElem(el);
+      // Prevent mouse clicks on link nodes from redirecting the browser tab.
+      el.addEventListener('click', (event) => {
```
jeremydw, line 100: Google AI mode suggests this is another way to handle this (untested, not sure if it works)

```
import { $isLinkNode } from '@lexical/link';
import { $getSelection, $isRangeSelection } from 'lexical';

editor.registerCommand(
  CLICK_COMMAND,
  (payload) => {
    const event = payload;
    const selection = $getSelection();

    if ($isRangeSelection(selection)) {
      const node = selection.getNodes()[0]; // Or iterate through nodes if multiple are selected
      if ($isLinkNode(node)) {
        event.preventDefault(); // Prevent default link behavior
        return true; // Mark as handled
      }
    }
    return false; // Not handled
  },
  COMMAND_PRIORITY_LOW,
);
```

packages/root-cms/ui/components/RichTextEditor/lexical/utils/convert-to-lexical.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+  }
+}
+
+function createNodesFromHTML(htmlString: string) {
+  const template = document.createElement('template');
+  template.innerHTML = htmlString;
+  const fragment = template.content;
+
+  const nodes: any[] = [];
```
jeremydw, line 61: nit: assign type?

packages/root-cms/ui/pages/SettingsPage/SettingsPage.tsx
```diff
@@ -37,6 +40,72 @@ export function SettingsPage() {
...
+                    rich text editor with the{' '}
+                    <a
+                      href="https://lexical.dev/"
+                      target="_blank"
+                      rel="noopener noreferrer nofollow"
+                    >
+                      Lexical
+                    </a>{' '}
+                    version. The data saved by the Lexical version is backwards
+                    compatible but the editor may be missing a few features.
```
jeremydw, line 91: suggestion: list the missing features

jeremydw (approved):

packages/root-cms/ui/components/RichTextEditor/lexical/LexicalEditor.tsx
stevenle: it's compatible but it requires a custom upload dialog from us which i haven't written yet.

packages/root-cms/ui/components/RichTextEditor/lexical/LexicalEditor.tsx
stevenle: done, realized similar logic already existed in `FloatingLinkEditorPlugin.tsx` and so i moved this logic there.

packages/root-cms/ui/pages/SettingsPage/SettingsPage.tsx
stevenle: i'd actually prefer to be more vague here, i think this is meant to be considered "experimental" for now, which may or may not launch if we find issues with this version of the editor.

PR #659, opened 2025-08-08
chore: add script for starting a cms dev server

stevenle:
The command `pnpm cms:dev` will now start a file watcher for the root-cms package, wait 2 seconds, then start the docs Root server.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "packageManager": "pnpm@8.9.0",
   "main": "index.js",
   "scripts": {
+    "cms:dev": "./scripts/dev_cms.sh",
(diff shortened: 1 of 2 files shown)


jeremydw:
does ctrl+c exit this cleanly? (i can never remember exactly how background processes work within sh scripts within package scripts)

stevenle:
yeah i see this when i ctrl+c:
```
[ui] npm run dev:signin exited with code SIGINT
--> Sending SIGTERM to other processes..
[core] npm run dev:core exited with code SIGINT
--> Sending SIGTERM to other processes..
[dev:ui] npm run dev:ui exited with code SIGINT
```

PR #660, opened 2025-08-08
fix: fix image non-empty state on load

stevenle:


jeremydw:
woops - thanks for fixing that

PR #661, opened 2025-08-08
feat: add site settings and google drive folder setting

jeremydw:
This PR adds the ability to manage site settings (settings associated with the site, not the user) in the CMS UI. It creates a framework for adding more settings over time.

For now, we just add one setting: "Google Drive folder". This allows site admins to specify a Google Drive folder that will be used by the localization process. If specified, when the localization process creates Google Sheets, it will place them in the Google Drive folder set in the site setting, rather than the invoker's "My Drive".

The settings are stored in Firestore, same as roles, meaning only admins can change them.

```
Projects/:project/settings/:key
```

image

image

image


stevenle:
when a new spreadsheet is created, the ACLs from the project are added to the new spreadsheet. when spreadsheets are created within a google drive folder, do you think we can skip that step (saves rpc time, plus sometimes it's buggy if a user account is deleted)

jeremydw:
ah yeah; that's a good thing to fix. how about this:

1. we could add a checkbox to the "create sheet" flow that says "Share Google Sheet with participants" and, if checked, will run through the current sharing flow
2. if the site has a Google Drive folder configured, that checkbox can be unchecked by default
3. if the site does _not_ have a Google Drive folder configured, that checkbox would remain checked

how about that?

stevenle:
checkbox sounds good to me!

jeremydw:
Made the following changes:

1. Added checkbox to toggle the sharing upon creation
2. Fixed a subtle bug with the "Open Google Sheet" button
3. Disabled the fields after the sheet is created

image

image

PR #669, opened 2025-08-14
fix: saving richtext data from edit flows

jeremydw:
@stevenle - Confirming this fixes the save flows but I'm not sure if it causes any performance issues. LMK what you think.


jeremydw:
Oh - it just occurred to me. Maybe we should add the ability to "force" an update, and leave the existing logic as-is? Then we could request a forced update from `updateItem` so that the time values are ignored.

We could implement the "force" in:

```
action.draft.updateKey(`${action.deepKey}.${key}`, newValue);
```

Of course, this adds new complex logic so the existing PR would be preferred (I think at least for now) if it doesn't cause a perf issue. But LMK what you think.

PR #671, opened 2025-08-14
fix: resize iframe when window resizes

jeremydw:
https://github.com/user-attachments/assets/5d2cec59-0121-47aa-a47a-e17f29224147

### DIFF
--- a/packages/root-cms/ui/pages/DocumentPage/DocumentPage.tsx
+++ b/packages/root-cms/ui/pages/DocumentPage/DocumentPage.tsx
@@ -22,7 +22,6 @@ import {UseDraftHook, useDraft} from '../../hooks/useDraft.js';
 import {useLocalStorage} from '../../hooks/useLocalStorage.js';
 import {Layout} from '../../layout/Layout.js';
 import {joinClassNames} from '../../utils/classes.js';
-import {debounce} from '../../utils/debounce.js';
 import {getDocPreviewPath, getDocServingPath} from '../../utils/doc-urls.js';
 
 interface DocumentPageProps {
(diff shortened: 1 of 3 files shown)


packages/root-cms/ui/pages/DocumentPage/DocumentPage.tsx
```diff
@@ -342,35 +342,48 @@ DocumentPage.Preview = (props: PreviewProps) => {
...
-    setIframeStyle({
-      '--iframe-width': `${width}px`,
-      '--iframe-height': `${height}px`,
-      '--iframe-scale': String(scale),
-    });
+    updateIframeStyle();
+
+    // Maintain the aspect ratio when the window is resized.
+    // Without this, the iframe gets cut off.
+    const handleResize = () => updateIframeStyle();
```
stevenle, line 381: can you debounce this?

stevenle: hmm let's go back to without the debounce, eric used to say css transitions create artificial latency and i'm feeling a little bit of that here. reason i suggested debouncing is to reduce cpu load since we have some perf problems at the moment anything we can do to help reduce unresponsiveness the better.

jeremydw: got it - yeah, there are tradeoffs with both. i do like how snappy and immediate it is _without_ the transitions so i'm aligned to that. i read a post the other day about how disabling transitions on macOS makes the system feel "faster" so i think this update is aligned to that feeling

stevenle (approved):

PR #695, opened 2025-08-17
feat: derive image preview if unspecified

jeremydw:
A week or two ago we added the ability for array items to specify image previews (to go along with their title previews).

- While this works, it's kind of tedious to have to specify this for every potential array item in the system, so instead, if the image preview is unspecified, we can derive the image preview by analyzing the fields and looking up the first image in the module data.
- As an added bonus we can more easily lookup the `canvasBgColor` too.
- Note: this only looks for images one level deep.

### DIFF
--- a/.changeset/blue-hotels-yawn.md
+++ b/.changeset/blue-hotels-yawn.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: derive image preview if unspecified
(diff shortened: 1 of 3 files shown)


stevenle:
chatted about this briefly the other day but i think this may be a no-go for a few reasons

(1) this feature can be computationally expensive, we just went through a major refactor to help reduce the amount of re-renders and fix ui jank, moving forward i need to "gatekeep" features that can slowly reintroduce issues over time. we don't want to have to walk the schema tree and data tree on every re-render for an image that may or may not exist on every single array item.

(2) ui changes that introduce "magic" that forces a developer to opt-out when they don't want it is an anti-pattern in my mind, it's easy enough for a developer to opt-in with the preview.image field and has greater clarity for both the ui as well as for the user.

PR #708, opened 2025-08-28
feat: add ability to customize the password page

jeremydw:
This PR adds the ability to customize the password page.

@stevenle Can you think of a better name or pattern to use to allow adding other components to the page? I called the option `children` but I'm not sure that's right.

FYI, I also swapped "Google Sans" for "Inter" since I don't think it's quite right to use Google's font for these pages.

Here's a screenshot of my example:

image


stevenle:
the inter change is good and we probably should've done this from the beginning.

re: customizing this page, i'm not in love with this pattern because it makes it difficult for us to change the underlying page style in the future without breaking the api (or having to create a prop for each new customization). ideally client would be able to fully define their own customized route like `routes/401.tsx` that this page automatically renders if it exists (client would just need create a `<form>` POST with a password field).

jeremydw:
> the inter change is good and we probably should've done this from the beginning.
> 
> re: customizing this page, i'm not in love with this pattern because it makes it difficult for us to change the underlying page style in the future without breaking the api (or having to create a prop for each new customization). ideally client would be able to fully define their own customized route like `routes/401.tsx` that this page automatically renders if it exists (client would just need create a `<form>` POST with a password field).

got it - i updated it and this works perfectly, but i'm seeking feedback on the following:

1. I added a new prop called `unauthorizedPage` to the plugin options that accepts a react component
2. In the blog example, I created a `routes/401.tsx` but I don't think this is quite right

It'd be good to establish the "right" pattern for sites to follow, maybe I should:

1. Move the `401.tsx` to `components/UnauthorizedPage/UnauthorizedPage.tsx`
2. Update the option name to something other than `unauthorizedPage` (or leave it as-is?)

Thoughts?

stevenle:
the issue is that root.config.ts is bundled separately from the rest of the root appplication, and so if you pass a component to  a plugin, it doesn't get rendered by the vite / root renderer (so for example style loading would be handled entirely differently than a normal component within a root app). this would be confusing to a developer trying to set up their component like any other component within their app.

when you create a page in `routes/401.tsx` (similar to `routes/404.tsx`), it gets picked up by the root router and the renderer can render it using something like `const [route] = router.get('/401')` and then `renderer.renderRoute(route)`. this would be handled by root renderer and would be able to pick up any css and other js deps as needed. i believe the renderer is injected to the req object, so you can do `const renderer = req.renderer`.

so i think the updates you'd need are:

- set up routes/401.tsx more similarly to routes/404.tsx (i.e. export a default component that is the "Page" component)
- when the password page needs to be rendered, use the router to check if a 401 page is defined (`req.renderer.router.get('/401')`) and if it is, use the `renderRoute()` function which i think should give you the html output where you can write it to the response (you can create a configuration option for this if you feel it's necessary, i'm fine with having it be "magic" with the 401)
- fallback to the default password page if the 401.tsx file isn't found

jeremydw:
yea; i thought that'd be a problem, although in my example it worked (although, admittedly i didn't try using CSS modules). i think this is a better change though.

one thing i wanted to do was to preserve the ability to customize the page displayed on a per-route basis, so i added an option on the route to support this (`unauthorizedPageUrl`). but, LMK if you want to rename/tune these. i also added `getRoute` to the `Renderer` to avoid needing to reconstruct the router.

packages/root-password-protect/src/plugin/plugin.tsx
```diff
@@ -172,10 +192,14 @@ async function verifySessionCookie(
...
   res: Response,
+  protectedRoute: PasswordProtectedRoute,
   props?: Omit<PasswordPageProps, 'nonce'>
 ) {
+  const [customRoute] =
+    req.renderer?.getRoute(protectedRoute.unauthorizedPageUrl || '/401') ?? [];
+  const Component = customRoute?.module.default ?? PasswordPage;
   const nonce = generateNonce();
-  const mainHtml = renderToString(<PasswordPage {...props} nonce={nonce} />);
+  const mainHtml = renderToString(<Component {...props} nonce={nonce} />);
```
stevenle, line 202: when using `req.renderer.getRoute()`, you'll need to use the renderer to render the html, otherwise if there are custom components (and other root-specific rendering logic) it wouldn't get picked up.

the renderToString() method here should only be used with the PasswordPage which renders the html directly using normal preact rendering (without the root magic).

stevenle:
> one thing i wanted to do was to preserve the ability to customize the page displayed on a per-route basis, so i added an option on the route to support this (`unauthorizedPageUrl`). but, LMK if you want to rename/tune these. i also added `getRoute` to the `Renderer` to avoid needing to reconstruct the router.

i think we can also send props to the route so that it's aware of the url it's rendering, so that you wouldn't have to define different routes for each password page. with the custom url tho you may want to pass it some prop so that the component knows that it's coming from a 401 so that it doesn't render 200 by mistake when crawlers and stuff fetch that route.

(it'll also be easier for things like sitemap to know what's a true route or not if we keep it to just the 401.tsx file)

jeremydw:
k - check out the latest; it's feeling a lot better now. but i see two potential problems:

1. the props aren't passed to the `renderRoute` function (just the `routeParams`):

```
      const data = await req.renderer?.renderRoute(route, {
        routeParams: {nonce},
      });
```

I think what we want to do is just have the router handle the route directly, but using the following won't work for this use case because the actual request won't match what the route is expecting (`/401`):

```
    return req.renderer?.handle(req, res, next);
```

2. (Maybe not a huge issue, but...) Because we delegate control of the rendering to the user-supplied route, it doesn't get the nonce/security headers unless the site opts into it in the root config.

Any suggestions for how to make what I have more idiomatic, or is it OK as is?

packages/root-password-protect/src/plugin/plugin.tsx
```diff
@@ -175,9 +182,26 @@ export async function renderPasswordPage(
...
-  res.send(html);
+  async function getHtml() {
+    // First, try to render the `/401` route. The route served at this path will
+    // display the "not authorized" interstitial page with a password prompt. The
+    // page should have a form on it that sends a POST request to itself with a
+    // form that contains a single field "password".
+    const route = req.renderer?.getRoute('/401')?.[0];
+    if (route) {
+      const data = await req.renderer?.renderRoute(route, {
+        routeParams: {nonce},
```
stevenle, line 196: we should reserve `routeParams` to be exclusively used for the url placeholder params. i'm thinking we should probably add some method to the renderer to handle how this is rendered and have the nonce logic handled by the root renderer.

(we'll probably also want to respect the `route.handle()` method, which is where we can pass any props to the component)

i can help with this more tomorrow when i have more time to think about this

packages/root-password-protect/src/plugin/plugin.tsx
```diff
@@ -107,7 +110,9 @@ async function handleProtectedRoute(
       !req.body.password
     ) {
       res.status(400);
-      renderPasswordPage(req, res, {error: 'Bad request (no password).'});
+      renderPasswordPage(req, res, {
+        error: 'Bad request (no password).',
```
stevenle, line 114: since we're exposing the password page, we may want to expose these errors as an exported enum/const so that developers can show an appropriate error message for the error.

packages/root-password-protect/src/plugin/plugin.tsx
jeremydw: yeah; sounds good. also, i realized i can just vendor the plugin into the project with the deadline and customize the page as a one-off, so we don't need to rush this for the time being. however, it'd probably be nice for sites to customize the password entry page so we should fix this up when time permits so sites have a canonical way of customizing it :)

stevenle:
created a child pr for the proposed changes above:
https://github.com/blinkk/rootjs/pull/711

take a look to see if this works on your end, i tested briefly and it works with vite's hmr so i think this will be a better developer experience.

PR #720, opened 2025-09-04
feat: flatten oneof types in collection schemas

stevenle:


stevenle:
@codex for some reason, running `pnpm build` in the `docs/` folder hangs. seems like an async function isn't properly resolving. can you help figure out if the change introduced any issues?

PR #796, opened 2025-11-06
chore: suppress verbatim undefined when using useTranslations

jeremydw:


packages/root/src/core/hooks/useTranslations.ts
```diff
@@ -26,6 +26,11 @@ export function useTranslations() {
   const stringParams = useStringParams();
   const middleware = useTranslationMiddleware();
   const t = (str: string, params?: Record<string, string | number>) => {
+    // Suppress verbatim `'undefined'` and `'null'` output, which can occur when
+    // using `useTranslations` with undefined values.
+    if (typeof str === 'undefined' || str === null) {
+      return undefined;
```
stevenle: i think if null is passed maybe null should be returned, so change `return undefined` to `return str`.

jeremydw: done; added comments

PR #804, opened 2025-11-12
feat: add stretch button

jeremydw:
This adds a stretch button to the CMS preview panel - the idea is to provide more vertical real estate for operators that are building really long pages, while preserving the desktop/tablet/mobile widths to help visualize the content at those widths.

https://github.com/user-attachments/assets/9947c04c-235a-41dd-8c46-504fa75a449a

### DIFF
--- a/packages/root-cms/ui/pages/DocumentPage/DocumentPage.tsx
+++ b/packages/root-cms/ui/pages/DocumentPage/DocumentPage.tsx
@@ -274,7 +274,7 @@ DocumentPage.Preview = (props: PreviewProps) => {
     'root::DocumentPage::preview::device',
     ''
   );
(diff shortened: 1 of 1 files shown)


jeremydw:
@stevenle - sounds good; i have a very content-heavy project this week so having the extra space immediately would be really helpful for the content entry for that. agree that nesting this inside a submenu (as a followup change with more features) sounds good.

FWIW, i do like showing the breakpoints clearly since it helps with discovery of the feature; but i think the "expand" feature could be nested inside a submenu with more options

stevenle:
> FWIW, i do like showing the breakpoints clearly since it helps with discovery of the feature; but i think the "expand" feature could be nested inside a submenu with more options

yeah i agree, we would keep the 3 device buttons but the submenu would appear below it when it contextually makes sense. a quick mockup:

image

PR #838, opened 2025-11-24
feat: add robust translations table

### LINKED ISSUE #400: Translation editor doesn't scale
For a site with a large number of locales, the translations editor doesn't seem to scale (the page stalls/freezes). Need a better system for handling large tabular data that doesn't crash the page.

jeremydw:
this PR adds ag-grid and replaces the translations table with it. some features:

- performance should be vastly improved for large translations databases
- improved UI that keeps the source and actions columns pinned
- add ability to filter by locale and tag
- add search field
- add ability to deep link to translations page with filters applied (so we can link to it from the localizer page)
- adds test (but, i couldn't figure out how to make the screenshot file larger)

note the CSS is inlined but i can have that extracted if needed

fixes #400


stevenle:
note that eventually when i'm able to overhaul the translations system, my plan is to have the ux feel similar to editing a csv file within a folder format, where you get an initial list of "folders" (collections) and you drill into a specific "file" (document) to edit and that presents you with the grid. within a grid view i'd like to be able to edit the cells directly and hit save (instead of opening another view). sharing this here for reference so that we can work eventually work toward how that would work.

jeremydw:
> note that eventually when i'm able to overhaul the translations system, my plan is to have the ux feel similar to editing a csv file within a folder format, where you get an initial list of "folders" (collections) and you drill into a specific "file" (document) to edit and that presents you with the grid. within a grid view i'd like to be able to edit the cells directly and hit save (instead of opening another view). sharing this here for reference so that we can work eventually work toward how that would work.

- the good news is ag-grid actually has a robust "edit mode" so with a few quick modifications it should be super easy to add editing abilities using this same component! once we get there we can add those features quite nicely i think.

- ag-grid also supports accordion-style elements in case we want a UX that allows you to drill-down into a specific "folder" (and not in a separate view or pane).

stevenle:
just pulled, yeah i think it's a little light, i may play around with darkening it across the board. i just flipped to the release page to compare, i think agree with the consistency comment.

the border around the search/filter elements probably should be updated to the same color for consistency as well.

also last thing i noticed before i merge is i think there's 2 levels of scrolling at the moment (table scrolls and the whole pane scrolls), is it easy to set a max-height on the table so that the pane stays under 100vh?

PR #853, opened 2025-12-03
feat: add import/export cli utility

### LINKED ISSUE #308: CLI command for backing up CMS data
**Description**

`root-cms backup outdir/` should store a backup of CMS data. Optionally can be limited to specific collections via a flag, e.g. `root-cms backup --collection=BlogPosts blog-backup/`.

jeremydw:
This utility aims to provide more ergonomic access to the built-in Firestore import/export utilities. The idea is to make it easy to transfer data in and out of the Root CMS Firestore (i.e. when transferring projects - or even just searching or modifying data locally).

- It supports options to override the GCP project ID, site ID, and database ID. 
- It supports an `--include` flag where you can specify Firestore collections (e.g. ActionLogs, Translations, Collections) etc. to include, and it recursively exports all files in those collections to a local dir
- Currently it exports all Root CMS collections (if we want to add the ability to limit the CMS collections I need to figure out a good way to differentiate from the Firestore collection and the Root CMS collection nomenclature on the CLI args)

Fixes #308

### DIFF
--- a/packages/root-cms/cli/cli.ts
+++ b/packages/root-cms/cli/cli.ts
@@ -44,14 +44,19 @@ class CliRunner {
       .description(
         'exports firestore data to a local directory\n\n' +
           'By default, all content in the Firestore database associated with the CMS will be exported.\n\n' +
+          'A unique new directory will be created for each export.\n\n' +
+          'File naming conventions:\n' +
+          "- Documents that act as containers for subcollections are exported as directories containing a `__data.json` file for the document's own data.\n" +
+          '- Standalone documents are exported as JSON files named after their document ID (e.g. `page.json`).\n' +
+          '- For collections like ActionLogs and Translations, the document ID (often a hash) is used as the filename.\n\n' +
+          'Usage examples:\n' +
+          '  $ root-cms export\n' +
+          '  $ root-cms export --filter "Collections/Pages/**"\n' +
+          '  $ root-cms export --filter "Collections/Pages/**,!Collections/Pages/Draft/**"\n\n' +
           'Example output:\n' +
           '  <output>/Collections/Pages/Draft/...\n' +
           '  <output>/Collections/Pages/Published/...\n' +
-          '  <output>/ActionLogs/...\n\n' +
-          'File naming conventions:\n' +
-          "- Documents that act as containers for subcollections are exported as a directory containing a `__data.json` file for the document's own data.\n" +
-          '- Standalone documents are exported as JSON files named after their document ID (e.g. `page.json`).\n' +
-          '- For collections like ActionLogs and Translations, the document ID (often a hash) is used as the filename.'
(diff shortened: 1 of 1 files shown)


packages/root-cms/cli/cli.ts
```diff
@@ -37,8 +39,37 @@ class CliRunner {
         'generates root-cms.d.ts from *.schema.ts files in the project'
       )
       .action(generateTypes);
+    program
+      .command('export')
+      .description('exports firestore data to a local directory')
+      .option(
+        '--include <types>',
+        'comma-separated list of collection types to export (ActionLogs, Collections, DataSources, Releases, Translations, Users)'
```
stevenle: if i wanted to export just a single collection, what does the command look like? similarly if i only wanted to export either draft or published content?

jeremydw: I originally intentionally didn't allow customizing that because it started to get confusing between Firestore collections and CMS collections, but I just refactored this to replace the "include" flag with a "filter" flag.

Now you can use commands like this to surgically export/import data, e.g

```
# Just draft content
root-cms export --filter=Collections/Pages/Draft/**

# Draft + published for Pages
root-cms export --filter=Collections/Pages/**

# No action logs
root-cms export --filter=!ActionLogs/**

# Everything (including the project doc / project settings file)
root-cms export
```

stevenle: Got it, so if I don't provide --filter, it'll organize everything into that folder format?

E.g.

```
<output>/Collections/Pages/Draft/...
<output>/Collections/Pages/Published/...
<output>/ActionLogs/... (how is this organized? by action type?)
```

Can you include those command examples somewhere in the CLI docs so that it appears when you run `root-cms export --help`?

jeremydw: Yeah; here's the output of the command now:

```
$ p exec root-cms export --help
Usage: root-cms export [options]

exports firestore data to a local directory

By default, all content in the Firestore database associated with the CMS will be exported.

Examples:
  <output dir>/Collections/Pages/Draft/...
  <output dir>/Collections/Pages/Published/...
  <output dir>/ActionLogs/...

File naming conventions:
- Documents that act as containers for subcollections are exported as a directory containing a `__data.json` file for the document's own data.
- Standalone documents are exported as JSON files named after their document ID (e.g. `my-doc.json`).
- For collections like ActionLogs and Translations, the document ID (often a hash) is used as the filename.

Options:
  --filter <pattern>       comma-separated list of glob patterns to filter content (e.g. Collections/Pages/**, !ActionLogs/**)
  --site <siteId>          site id to export (overrides root config)
  --database <databaseId>  firestore database id (overrides root config, default: "(default)")
  --project <projectId>    gcp project id (overrides root config)
  -h, --help               display help for command
```

PR #855, opened 2025-12-03
feat: add ability to specify timezone on datetimefield

### LINKED ISSUE #854: Enhance timezone support for date and datetime fields
- Allow users to pick the timezone upon content entry.
- Allow developers to limit the field to a specific timezone of their choosing.

jeremydw:
This adds the ability to change the timezone 

- This doesn't change the underlying data stored in any way.
- If the user picks a different time zone in the UI, the value updates to reflect that. It defaults to the user's system timezone on reload.
- If the schema is fixed to a specific time zone, then the user can't change it in the UI.

image

Fixes #854

### DIFF
--- a/packages/root-cms/ui/components/DocEditor/fields/DateTimeField.tsx
+++ b/packages/root-cms/ui/components/DocEditor/fields/DateTimeField.tsx
@@ -77,7 +77,7 @@ export function DateTimeField(props: FieldProps) {
 
   return (
     <div className="DocEditor__DateTimeField">
-      <div style={{display: 'flex', gap: '8px', alignItems: 'center'}}>
+      <div className="DocEditor__DateTimeField__row">
         <input
           type="datetime-local"
           value={dateStr}
@@ -85,13 +85,10 @@ export function DateTimeField(props: FieldProps) {
             const target = e.target as HTMLInputElement;
             onDateChange(target.value, target.validity);
           }}
-          style={{flex: 1}}
+          className="DocEditor__DateTimeField__input"
         />
         {field.timezone ? (
-          <div
-            className="DocEditor__DateTimeField__timezone"
-            style={{marginTop: 0}}
-          >
+          <div className="DocEditor__DateTimeField__timezone">
             timezone: {field.timezone}
           </div>
         ) : (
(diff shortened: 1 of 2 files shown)


packages/root-cms/core/schema.ts
```diff
@@ -57,10 +76,16 @@ export function date(field: Omit<DateField, 'type'>): DateField {
 export type DateTimeField = CommonFieldProps & {
   type: 'datetime';
   default?: string;
+  /**
+   * Timezone to use for the datetime field. If set, the field will be limited to
+   * this timezone.
```
stevenle: Provide an example of a correct tz value in the jsdoc, e.g. "America/Los_Angeles".

packages/root-cms/ui/components/DocEditor/fields/DateTimeField.tsx
```diff
@@ -9,77 +9,127 @@ import {FieldProps} from './FieldProps.js';
 export function DateTimeField(props: FieldProps) {
   const field = props.field as DateTimeFieldSchema;
   const [value, setValue] = useDraftDocValue<Timestamp | null>(props.deepKey);
+
+  // Metadata is stored in a sibling field prefixed with `@`.
+  // e.g. if field is `scheduledAt`, metadata is in `@scheduledAt`.
+  const metadataKey = getMetadataKey(props.deepKey);
+  const [metadata, setMetadata] = useDraftDocValue<Record<string, any>>(
```
stevenle: For now since we only need the `timezone` metadata key, we should listen on only that key instead of the whole metadata object.

packages/root-cms/ui/components/DocEditor/fields/DateTimeField.tsx
```diff
@@ -1,55 +1,126 @@
...
-        value={dateStr}
-        onChange={(e: Event) => {
-          const target = e.target as HTMLInputElement;
-          const newDateStr = target.value;
-          onChange(newDateStr);
-        }}
-      />
-      <div className="DocEditor__DateTimeField__timezone">
-        timezone: {Intl.DateTimeFormat().resolvedOptions().timeZone}
+      <div style={{display: 'flex', gap: '8px', alignItems: 'center'}}>
```
stevenle: Can you move these inline styles to css?

jeremydw: done

jeremydw:
> Screenshot 2025-12-04 at 6 35 19 PM
> actually one other thing - for narrower widths, can you have the timezone drop to the next line?

done + done

PR #870, opened 2025-12-16
feat: add task manager db calls

stevenle:
## Summary
- add Firestore utilities for creating tasks and tracking assignee updates
- support default task assignee storage in site settings and utilities for setting it
- add task comment APIs that preserve edit and delete history

## Testing
- not run (not requested)

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_6941999bc5f0832384c0069d7049b8ff)

### DIFF
--- a/.changeset/ninety-panthers-kick.md
+++ b/.changeset/ninety-panthers-kick.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: add task manager db calls (#870)


jeremydw:
LGTM - we might also want to add a status and CC fields.

BTW, I considered suggesting naming these "Issues" instead of "Tasks" but I think the term "Tasks" aligns better with our vision for how this will eventually be used. (GitHub Issues and Google Buganizer call them "Issues" hence the thought.)

stevenle:
added status field, i'll probably refine this more once i start working on the ui (which i'll probably do by hand). cc field will need to come a little later when i figure out the email api, right now the tasks don't have a notification system so i'll need to figure that part out separately.

PR #880, opened 2026-01-04
chore: update the canvas bg logic

jeremydw:
- adds a UI allowing the user to toggle the dark background
- fixes a subtle bug with the file field menu overlapping array items

image

image


jeremydw:
What do you think of this? (I changed the opacity too to make it clearer that it's unchecked.) Not sure it's perfect but not sure what to do. :\

image

image

stevenle (approved):

PR #883, opened 2026-01-12
feat: add the ability to paste from gdrive links

jeremydw:
This PR adds the ability to paste text (Google Drive URLs - e.g. `https://drive.google.com/open?id=1q4466dGkfhb9vKArGxopAkqB3QcjnjQn&usp=drive_copy`) into the file field (when focused) in order to have Root CMS download the file from gDrive then upload it to the CMS.

### DIFF
--- a/packages/root-cms/ui/hooks/useGapiClient.ts
+++ b/packages/root-cms/ui/hooks/useGapiClient.ts
@@ -51,9 +51,9 @@ export function useGapiClient(): GapiClient {
   function login(options?: {scopes?: string[]}) {
     return new Promise<void>((resolve) => {
       const clientId = window.__ROOT_CTX.gapi!.clientId!;
-      const scopes = [...SCOPES, ...(options?.scopes || [])];
-      // De-dupe scopes.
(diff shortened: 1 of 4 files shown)


packages/root-cms/ui/hooks/useGapiClient.ts
```diff
@@ -47,12 +48,15 @@ export function useGapiClient(): GapiClient {
     }
   }, []);
 
-  function login() {
+  function login(options?: {scopes?: string[]}) {
     return new Promise<void>((resolve) => {
       const clientId = window.__ROOT_CTX.gapi!.clientId!;
+      const scopes = [...SCOPES, ...(options?.scopes || [])];
+      // De-dupe scopes.
+      const uniqueScopes = [...new Set(scopes)];
```
stevenle: I think I prefer to use `Array.from(new Set(scopes))` instead of the spread operator. Probably better to do it in one line as well, e.g.

```
const scopes = Array.from(new Set([
  ...SCOPES,
  ...(options?.scopes || []),
]));
```

PR #891, opened 2026-01-29
feat: add publish message support

jeremydw:
This PR adds "publish messages" which are the equivalent of commit messages.

- Adds a textarea on the publish dialog
- Adds ability to AI generate the publish message
- Publish message is attached to the version file
- Publish message is displayed in the version history for later auditing


packages/root-cms/ui/utils/doc.ts
```diff
@@ -64,6 +64,7 @@ export interface CMSDoc {
 export type Version = CMSDoc & {
   _versionId: string;
   tags?: string[];
+  message?: string;
```
jeremydw: @stevenle Is this OK or does it need to be nested or placed somewhere else?

jeremydw:
yup, updated to "commit message"

stevenle (approved):

PR #896, opened 2026-01-30
feat: add translation metadata and quick ai translations

jeremydw:
This PR adds two major features:

1. Metadata on translatable fields
2. Quick AI translations

The current metadata options are:

1. Description - Included in translation requests (in ARB files and AI translations)
2. "Do Not Translate" checkbox. If this is checked, the string isn't extracted or sent for translation.

The metadata is stored as:

```
{
  fieldName: 'Value',
  @fieldName: {
    translations: {
      doNotTranslate: true,
      description: 'Foo bar'
    }
  }
}
```

When "do not translate" is checked the icon in the doc editor switches to red with a strike through it.

image

image

image

image

### DIFF
--- a/packages/root-cms/ui/components/EditTranslationsModal/EditTranslationsModal.tsx
+++ b/packages/root-cms/ui/components/EditTranslationsModal/EditTranslationsModal.tsx
@@ -81,10 +81,10 @@ export function EditTranslationsModal(
   const [description, setDescription] = useState('');
   const [aiGenerating, setAiGenerating] = useState(false);
 
-  // Use draft from props (passed from DocEditor)
+  // Use draft from props (passed from DocEditor).
   const draft = props.draft || null;
 
-  // Load translation metadata from the document
+  // Load translation metadata from the document.
   useEffect(() => {
     if (props.field?.deepKey && draft?.controller) {
       const metadataKey = getMetadataKey(props.field.deepKey);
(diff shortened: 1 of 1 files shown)


packages/root-cms/core/ai.ts
```diff
@@ -437,3 +437,102 @@ export class ChatClient {
...
+
+  if (options.description) {
+    userPrompt += `Context/Description: ${options.description}\n\n`;
+  }
+
+  if (
+    options.existingTranslations &&
+    Object.keys(options.existingTranslations).length > 0
+  ) {
+    userPrompt += 'Existing translations for reference:\n';
```
stevenle: loop-based string concatentation is expensive, use arrays for building strings

packages/root-cms/core/ai.ts
```diff
@@ -437,3 +437,102 @@ export class ChatClient {
...
+      },
+    ],
+  });
+
+  const responseText = res.text || '{}';
+
+  // Try to extract JSON from the response
+  let jsonText = responseText.trim();
+
+  // Remove markdown code blocks if present
```
stevenle: This might be worth extracting to its own function and adding unit tests for it.

packages/root-cms/core/ai.ts
```diff
@@ -437,3 +437,102 @@ export class ChatClient {
...
+
+  const systemPrompt = [
+    'You are a professional translator assistant.',
+    'Translate the given source text into the requested target languages.',
+    'Maintain the tone, style, and intent of the original text.',
+    'Return ONLY a valid JSON object with locale codes as keys and translations as values.',
+    'Do not include any markdown formatting, code blocks, or explanatory text.',
+  ].join('\n');
+
+  let userPrompt = `Source text: "${options.sourceText}"\n\n`;
```
stevenle: loop-based string concatentation is expensive, use arrays for building strings

PR #898, opened 2026-02-02
feat: implement schema.glob function for schema pattern resolution

jeremydw:
- We've had some use cases where the manual import.meta.glob convention is hard to remember, plus that pattern isn't compatible with circular references or nesting the same schema within itself.

- So,  add `allSchemas` as a new function provided by `schema.ts` that supports circular schema references and cleans up duplicate code.

- Add a full example within `TemplateSandbox`.

image

image

### DIFF
--- a/.changeset/stupid-olives-lick.md
+++ b/.changeset/stupid-olives-lick.md
@@ -2,4 +2,4 @@
 '@blinkk/root-cms': patch
 ---
 
-feat: implement allSchemas for schema pattern resolution
+feat: implement glob for schema pattern resolution
--- a/packages/root-cms/core/project.integration.test.ts
+++ b/packages/root-cms/core/project.integration.test.ts
@@ -99,13 +99,13 @@ describe('Production Bundle Integration', () => {
 });
 
 describe('SchemaPattern Resolution', () => {
-  it('should export allSchemas function from core module', async () => {
+  it('should export glob function from core module', async () => {
     const coreModule = await import('./core.js');
 
     expect(typeof coreModule.schema.glob).toBe('function');
   });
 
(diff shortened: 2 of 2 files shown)


stevenle:
thanks this looks great, i'm just thinking about the naming. did you consider any other alternatives? `schema.allSchemas()` feels redundant and is a little bit of a misnomer because i think you should always be required to pass it a path or pattern.

jeremydw:
Switched it to `schema.glob`

packages/root-cms/core/schema.ts
```diff
@@ -428,3 +451,63 @@ export function defineCollection(
...
+ *         types: schema.glob('/templates/*\/*.schema.ts'),
+ *       }),
+ *     }),
+ *   ],
+ * });
+ *
+ * // With exclusions:
+ * schema.oneOf({
+ *   types: schema.glob('/templates/*\/*.schema.ts', {
+ *     exclude: ['DeprecatedTemplate'],
```
stevenle: nothing to change here, but i just want to note that there may be future problems with this api interface.

1) user may want to be able to glob multiple dirs similar to the way import.meta.glob accepts lists (including negative globs), so maybe in the future we should support string arrays (no need to change it now, let's keep the interface simple)
2) it's possible two files export the same schema name which will cause collisions with the `exclude` option. in general i consider it an antipattern to have colliding schema names, but we may have a scenario where a user pulls templates from both `packages/designsystem/templates` and `site/templates` and there may be overrides in the latter.

user can work around this issue by looping/filtering so this isn't a concern for me at the moment, just noting that we may want to extend the functionality in the future.

PR #904, opened 2026-02-06
feat: add button to apply tag to doc's strings

### LINKED ISSUE #774: Button for syncing translations to a doc
**Description**

Sometimes a doc uses strings/translations that aren't tagged to a doc. We should have a button somewhere that extracts all the translateable strings, finds all the DB entries for the translations, and properly tags the doc for those translations.

stevenle:
Sometimes translations are imported via doc A and the same strings are used on doc B, but since those translations aren't tagged for doc B, the translations aren't loaded when the page renders.

This change gives content editors a quick way to "sync" all translations available in the system by applying the doc's tags for all strings extracted from the doc.

Screenshot 2026-02-12 at 8 18 24 PM

Screenshot 2026-02-12 at 8 18 30 PM

fixes #774

### DIFF
--- a/.changeset/curly-eyes-design.md
+++ b/.changeset/curly-eyes-design.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: add button for appending tag to a doc's string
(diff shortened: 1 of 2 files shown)


stevenle:
@jeremydw this is ready for another look. per chat, ui is updated to be more of a warning box than a permanent button. updated the PR description with the latest screenshots.

jeremydw:
UI looks good 👍

PR #906, opened 2026-02-10
chore: add safeguards when using setRawDoc

jeremydw:
I keep accidentally malforming timestamps when writing scripts that use `setRawDoc` which prevents the Root CMS UI from working. This adds some basic safeguards when using `setRawDoc` to the structure/type of the Root CMS system fields to help avoid that.

### DIFF
--- a/packages/root-cms/core/client.ts
+++ b/packages/root-cms/core/client.ts
@@ -8,6 +8,7 @@ import {
   Timestamp,
   WriteBatch,
 } from 'firebase-admin/firestore';
+import {normalizeSlug} from '../shared/slug.js';
 import {CMSPlugin} from './plugin.js';
 import {Collection} from './schema.js';
 import {TranslationsManager} from './translations-manager.js';
@@ -262,7 +263,7 @@ export class RootCMSClient {
 
     const modeCollection = this.getModeCollection(options.mode);
     // Slugs with slashes are encoded as `--` in the DB.
-    slug = slug.replaceAll('/', '--');
+    slug = normalizeSlug(slug);
     const dbPath = `Projects/${this.projectId}/Collections/${collectionId}/${modeCollection}/${slug}`;
     const docRef = this.db.doc(dbPath);
     const doc = await docRef.get();
@@ -300,7 +301,7 @@ export class RootCMSClient {
   ) {
     const collectionDocsPath = this.dbCollectionDocsPath(collectionId, options);
     // Slugs with slashes are encoded as `--` in the DB.
-    const normalizedSlug = slug.replaceAll('/', '--');
+    const normalizedSlug = normalizeSlug(slug);
     return `${collectionDocsPath}/${normalizedSlug}`;
   }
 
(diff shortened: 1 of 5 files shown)


packages/root-cms/core/client.ts
```diff
@@ -1467,6 +1532,122 @@ export function unmarshalData(data: any): any {
...
+  }
+
+  // Validate publishingLocked if present.
+  if (sys.publishingLocked !== undefined && sys.publishingLocked !== null) {
+    if (typeof sys.publishingLocked !== 'object') {
+      throw new Error('Invalid sys.publishingLocked: expected object or null');
+    }
+    if (sys.publishingLocked.until !== undefined) {
+      const until = sys.publishingLocked.until;
+      if (
```
stevenle: this timestamp conversion logic seems to be duplicated here and above, maybe move the logic to a normalization function, e.g.

```
until: convertToTimetamp(until)
```

packages/root-cms/core/client.ts
```diff
@@ -1467,6 +1532,122 @@ export function unmarshalData(data: any): any {
...
+      else if (typeof val === 'number') {
+        result[field] = Timestamp.fromMillis(val);
+      }
+      // If it's a Date, convert to Timestamp.
+      else if (val instanceof Date) {
+        result[field] = Timestamp.fromDate(val);
+      }
+      // Otherwise, it's invalid.
+      else {
+        throw new Error(
```
stevenle: this error might catch callers off guard, because when creating a new doc you should be able to save only the fields and the system should automatically set a few default sys values if it's not provided.

- if createdAt and modifiedAt are undefined, it should be set to the current time
- if createdBy and modifiedBy are undefined, set it to the current user (i don't think there's currently a way to figure out "who" the current user is, so maybe default it to `root-cms-client` if not specified via a argument option)
- if locales is empty, set it to [\<default locale\> or 'en']

packages/root-cms/core/client.ts
```diff
@@ -400,15 +400,80 @@ export class RootCMSClient {
...
+   *     modifiedBy: 'user@example.com',
+   *     locales: ['en', 'es']
+   *   },
+   *   fields: {
+   *     title: 'Home Page'
+   *   }
+   * }, { mode: 'draft' });
+   * ```
    */
   async setRawDoc(
```
stevenle: one other validation that should be added to this is to ensure the slug/collection/id fields are correct in the updated doc.

there are cases where we might to something like:

```
setRawDoc('Pages', 'bar', getRawDoc('Pages', 'foo'))
```

in this case a user would be incorrectly setting the new doc to `id = 'Pages/foo'` which is incorrect

jeremydw:
done - updated all the feedback. LMK if you'd prefer different defaults or different comments!

packages/root-cms/core/client.ts
```diff
@@ -400,17 +413,111 @@ export class RootCMSClient {
...
+
+    // Ensure id, collection, and slug fields match the parameters to prevent data inconsistencies.
+    const expectedId = `${collectionId}/${slug}`;
+    data.id = expectedId;
+    data.collection = collectionId;
+    data.slug = slug;
+
+    // Validate and normalize sys fields to prevent data integrity issues.
+    // Default to empty object if sys is not provided.
+    if (!data.sys) {
```
stevenle: remove this if() statement and update to `data.sys = validateSysFields(data.sys || {})` below

PR #997, opened 2026-03-27
feat: add docs.update to cli and fix schema resolution

### LINKED ISSUE #913: saveDraftData with validate: true throws 'glob is not a function' error in Node.js scripts
## Description

When using `RootCMSClient.saveDraftData()` with `validate: true` in a Node.js script, an error is thrown:

```
TypeError: (intermediate value).glob is not a function
    at file:///node_modules/@blinkk/root-cms/dist/chunk-RNDSZKAW.js:2:34
    at RootCMSClient.getCollection
    at RootCMSClient.saveDraftData
```

## Root Cause

The issue is in `core/project.ts` which uses `import.meta.glob()`:

```typescript
var SCHEMA_MODULES = import.meta.glob(
  [
    "/**/*.schema.ts",
    "!/appengine/**/*.schema.ts",
    ...
  ],
  { eager: true }
);
```

`import.meta.glob` is a **Vite-specific feature** that doesn't exist in standard Node.js. When running scripts outside of Vite (e.g., migration scripts, CLI tools), this fails.

## Steps to Reproduce

1. Create a Node.js script that uses RootCMSClient
2. Call `saveDraftData` with `validate: true`:

```javascript
import {RootCMSClient} from '@blinkk/root-cms';
import {loadRootConfig} from '@blinkk/root/node';

const rootConfig = await loadRootConfig(rootDir, {command: 'dev'});
const cmsClient = new RootCMSClient(rootConfig);

await cmsClient.saveDraftData('Pages/research', fields, {
  modifiedBy: 'script',
  validate: true,  // This causes the error
});
```

## Recommended Fix

**Option 1: Runtime detection with fallback**

```typescript
// In core/project.ts
let SCHEMA_MODULES = {};

if (typeof import.meta.glob === 'function') {
  // Vite environment
  SCHEMA_MODULES = import.meta.glob([...], { eager: true });
} else {
  // Node.js fallback - use fs to find and import schemas
  // This could be initialized lazily when getProjectSchemas() is called
}
```

**Option 2: Separate entry points**

Create separate bundles for browser/Vite vs Node.js environments, with the Node.js version using `fs.readdirSync` + dynamic `import()` instead of `import.meta.glob`.

**Option 3: Pass schemas via config**

Allow the root config to include pre-loaded schemas when running in Node.js:

```typescript
const rootConfig = await loadRootConfig(rootDir, {
  command: 'dev',
  preloadSchemas: true  // Would load schemas using fs
});
```

## Environment

- root-cms version: 2.5.1
- Node.js version: v22.x

## Workaround

Using `saveDraftData` without `validate: true` works. Using `setRawDoc` also works.

jeremydw:
This PR adds a `docs.update` to the CLI and fixes the schema resolution bug I introduced with `schema.glob`

- Add `docs.update <docId> <fieldPath> [value]` CLI command for surgically updating a single field in a draft doc by its deep key path
- Support inline JSON values, --file, and stdin for providing the value
- Validate against collection schema by default (--no-validate to skip)
- Add `resolveFieldAtPath()` to reject unknown field paths during validation (e.g. `foo.bar` on a schema without a `foo` field)
- Fix `import.meta.glob is not a function` error in CLI context by loading schemas via `viteSsrLoadModule` and passing them through a new `collectionSchema` option on SaveDraftOptions/UpdateDraftOptions
- Add optional `collectionSchema` to SaveDraftOptions and UpdateDraftOptions to bypass Vite-dependent `getCollection()` call

fixes #913


stevenle:
i think we should add some consistency between `docs.update`, `docs.set`, and `docs.get`. if we choose to add `--file` flag to `docs.update`, we should update the latter two to do the same. and if `docs.update` accepts the value as a command line arg, the `docs.set` should probably do the same.

can you (or ai) take a stab at what all 3 interfaces should look like?

PR #1005, opened 2026-03-29
feat: add native root jsx runtime and renderer

### LINKED ISSUE #864: Native TSX rendering
**Description**

The root rendering system currently depends on the client to install preact in their package or monorepo, this is due to some limitation of preact's context system that requires a singleton, and if there are multiple preact installations an error occurs. This causes issues with root native components that rely on context.

We should look into having root render tsx directly and providing minimal hooks that works for server side rendering.

stevenle:
## Summary
This PR replaces the external `preact-render-to-string` dependency with a custom built-in JSX-to-HTML renderer implementation. The new renderer provides equivalent functionality while reducing external dependencies and enabling better control over rendering behavior.

## Key Changes
- **New JSX renderer** (`jsx-render.ts`): Implements a complete Preact VNode-to-HTML string renderer with support for:
  - HTML elements, void elements, and fragments
  - Functional and class components
  - Context providers and consumers with `useContext` hook support
  - Proper HTML entity escaping in text content and attributes
  - Style object-to-string conversion with camelCase-to-kebab-case conversion
  - JSX prop name to HTML attribute name mapping (e.g., `className` → `class`)
  - Event handler filtering
  - `dangerouslySetInnerHTML` support

- **Rendering modes**: 
  - `'pretty'` mode (default): Block-level elements render on their own lines with newlines
  - `'minimal'` mode: Compact output with no extra whitespace
  - Configurable custom block elements via `blockElements` option

- **Configuration**: Added `jsxRenderer` config option to `RootUserConfig` for controlling renderer behavior

- **Integration**: Updated `Renderer` class to use the new `renderJsxToString` function instead of `preact-render-to-string`

- **Dependency removal**: Removed `preact-render-to-string` from peer dependencies in `package.json`

- **Test updates**: Updated snapshot tests to reflect minor formatting differences (void elements no longer have self-closing slashes, newline handling in pretty mode)

## Implementation Details
- The renderer properly handles Preact's hook system by creating fake component instances with the necessary internal properties (`__v`, `__d`, `__h`, `__s`, `__H`)
- Context stacks are maintained during rendering to support nested providers
- Preact option hooks (`__b`, `__r`, `diffed`) are called to integrate with the hooks addon
- HTML escaping is optimized to only escape when necessary (& < > in text; & " < > in attributes)

https://claude.ai/code/session_013BGyEWjQUa4yKAdv4zRKiZ

fixes #864


jeremydw:
Looks awesome; looking forward to having more control over the SSG output as well as potential performance improvements.

Two quick thoughts:

1. Can projects opt-into this via a config in tsconfig.json or a flag elsewhere? It'd be good to have a migration period where we can test the different runtime and more easily flip back and forth with a config change if needed. For example, we might want to first opt in our SSG sites to this and/or lower-visibility SSR sites for a time period before more broadly enabling it.

2. I saw some hardcoded lists of HTML element types in `packages/root/src/jsx/jsx-render.ts`. Is this often updated in Preact? We may want to establish a pattern for subscribing to changes so we can keep ours up to date as well.

PR #1028, opened 2026-04-10
feat: add move doc action

jeremydw:
- a few folks recently have asked how to move docs and i have to explain the "copy + delete" flow
- let's add an explicit move doc feature so that this can be done in a single action

### DIFF
--- a/packages/root-cms/ui/components/DocActionsMenu/DocActionsMenu.tsx
+++ b/packages/root-cms/ui/components/DocActionsMenu/DocActionsMenu.tsx
@@ -7,6 +7,7 @@ import {
   IconCloudOff,
   IconCopy,
   IconDotsVertical,
+  IconFileExport,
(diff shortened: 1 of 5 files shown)


stevenle:
i think we've had this conversation in the past, but i consider "move" a potentially destructive action which i think we should limit options for. per moore's law anything that can go wrong will go wrong.

reasons:

- if a live page is "moved", the old url will immediately stop working, not all users will understand that there should be a migration / redirect plan
- there will be references to the old doc in reference fields and releases that would need to be updated to point to the new reference, this would be an expensive operation that has many touchpoints
- it's not that many clicks to "copy" instead (and then delete the old page) which is less destructive and users can be easily "taught" how to handle this flow without breaking any existing pages

i would consider this if the alternatives weren't straightfowrard but it's easy enough to just use the "copy" feature.

PR #1032, opened 2026-04-14
feat: add ability to archive docs

stevenle:
## Summary
This PR adds the ability to archive and unarchive documents in the CMS. Archived documents are hidden from collection list views by default but can be restored later, providing a softer alternative to permanent deletion.

## Key Changes
- **Archive/Unarchive Actions**: Added `cmsArchiveDoc()` and `cmsUnarchiveDoc()` functions to manage document archive state by setting/clearing `sys.archivedAt` and `sys.archivedBy` fields
- **UI Controls**: 
  - Added Archive/Unarchive menu items in `DocActionsMenu` with confirmation dialog for archiving
  - Added "Show archived" toggle switch in `CollectionPage` to control visibility of archived documents
  - Added archive status badge in `DocStatusBadges` showing when and by whom a document was archived
- **Filtering Logic**: Updated `useDocsList` hook to filter out archived documents by default, with an `includeArchived` option to show them when the toggle is enabled
- **State Tracking**: Extended `CMSDoc` interface to include `archivedAt` timestamp and `archivedBy` email fields
- **Action Events**: Added 'archive' and 'unarchive' action types to `DocActionEvent` for proper event handling

## Implementation Details
- Archive state is persisted in Firestore under `sys.archivedAt` and `sys.archivedBy`
- Archived documents remain fully intact and cached; they're only hidden from list views
- The archive toggle state is persisted to localStorage per collection for user preference
- All archive/unarchive operations are logged via the existing `logAction()` utility

https://claude.ai/code/session_01LZBF4kWyaz1eNVWSHLTkzy

### DIFF
--- a/.changeset/nervous-peas-know.md
+++ b/.changeset/nervous-peas-know.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: add ability to archive docs #1032
--- a/packages/root-cms/ui/components/DocActionsMenu/DocActionsMenu.tsx
+++ b/packages/root-cms/ui/components/DocActionsMenu/DocActionsMenu.tsx
@@ -439,13 +439,15 @@ export function DocActionsMenu(props: DocActionsMenuProps) {
           Unarchive
         </Menu.Item>
       ) : (
-        <Menu.Item
-          icon={<IconArchive size={20} />}
-          onClick={() => onArchiveDoc()}
-          disabled={!canEdit}
-        >
-          Archive
-        </Menu.Item>
+        !sys.publishedAt && (
+          <Menu.Item
+            icon={<IconArchive size={20} />}
+            onClick={() => onArchiveDoc()}
+            disabled={!canEdit}
+          >
+            Archive
+          </Menu.Item>
+        )
       )}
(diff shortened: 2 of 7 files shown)


jeremydw (commented):
Do we need to make changes to the UX when a document is archived? I was thinking archived documents need to be treated as read only (e.g. no ability to edit them, no publishing or scheduling).

I think we also may want to show an "Archived" badge as the status for any reference fields that might still point to the doc. (Other CMS do this for archived items).

stevenle:
these are good points, let me see if claude can easily update these without too much hassle. i may handle the "read only" version in a separate PR since i think that would be more involved. i was thinking in the short term it would just remove it from the collection list and that would be "good enough" (at least for my use case, which is managing a list of experiments that go stale once the experiment ends)

stevenle:
Screenshot 2026-04-14 at 7 34 48 PM

jeremydw:
LGTM, we should also add an "Archived" status badge (if we don't have one already)

stevenle:
yup it's there:

Screenshot 2026-04-14 at 7 40 30 PM

i still need to verify the ui before merging, will merge tomorrow.

PR #1052, opened 2026-04-24
feat: add email sending service, rename gci to root-services

stevenle:


stevenle:
@jeremydw I separated out the backend changes from the other PR, the way this works is that it reads firestore for pending emails and sends them. A client wanting to send emails would simply save an entry to firestore.

This separation helps us avoid having to write an authentication layer since presumably any caller that is able to write to firestore should be able to send emails. Someone hitting the endpoint should not have any ability to read the contents of the email whatsoever.

Emails by default should have a 1hr expiration so that if there are any issues and emails get backed up, we don't end up sending thousands of emails by mistake.

jeremydw (approved):

PR #1065, opened 2026-04-28
feat: add task manager ui

### LINKED ISSUE #857: Task manager on CMS home page
**Description**

The CMS home page should be redesigned to feature a task manager that is capable of agentic workflows. When the user navigates to the /cms/ URL it should be presented with:

- A prompt-style text box that asks something like "What do you want to do today?"
- A list of open tasks
- A sidebar of recent actions (from the action log)

When a user fills out the text box, it should create a new task that is capable of having an AI agent make updates. An AI-assigned task should present "diffs" of proposed changes and the user should be able to iterate on those changes. There should be a preview link for any proposed changes by the AI agent.

Note: "tasks" and "reviews" should be two different data structures, tasks should be generic (like bug tracking or new feature requests) and not necessarily be tied to AI or reviews.

stevenle:
Note this is an MVP of the task manager UI, we'll build out more features as time goes on. Need to get this in before I can start prototyping the "assign to AI" features.

### Screenshots:

Screenshot 2026-04-30 at 8 33 15 AM

Screenshot 2026-04-30 at 8 24 46 AM

Screenshot 2026-04-30 at 8 24 43 AM

Screenshot 2026-04-30 at 8 32 59 AM

### Referenced Issues

Fixes #857

### DIFF
--- a/.changeset/hungry-suns-live.md
+++ b/.changeset/hungry-suns-live.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: add task manager ui
--- a/packages/root-cms/core/plugin.ts
+++ b/packages/root-cms/core/plugin.ts
@@ -53,6 +53,7 @@ export type {
 export type CMSBuiltInSidebarTool =
   | 'home'
   | 'content'
+  | 'tasks'
   | 'releases'
   | 'data'
   | 'assets'
--- a/packages/root-cms/ui/components/RichTextEditor/lexical/LexicalEditor.css
+++ b/packages/root-cms/ui/components/RichTextEditor/lexical/LexicalEditor.css
@@ -20,6 +20,18 @@
   z-index: 1;
 }
 
+.LexicalEditor--comment .LexicalEditor__toolbar {
+  border-radius: 6px 6px 0 0;
+  gap: 6px;
+  margin-bottom: -1px;
+  padding: 6px;
+  position: static;
+}
+
+.LexicalEditor--comment .LexicalEditor__toolbar--comment {
+  flex-wrap: wrap;
+}
+
 .LexicalEditor__toolbar__dropdown {
   height: 28px;
   font-size: 12px;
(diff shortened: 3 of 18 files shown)


stevenle:
@jeremydw this is nowhere near "feature complete" but i ran out of tokens and am eager to "launch and iterate" this feature. let me know if at a glance you see any glaring issues.

some missing features:
- file attachments
- image uploads / screenshots (it would be cool to add an annotation tool)
- advanced features like marking ticket as duplicate, tags/labels, etc.
- assign to ai (!)

one thing i wanted to ask is about the statuses - i'm thinking we may want a "work done, need review approval" state. the "closed" state is also something i'm trying to figure out - on one hand, i don't want a bunch of different states but i want to be able to distinguish between "closed - job is done and launched" vs "closed - will not do" vs. "closed - some other reason". one thing that we may also consider is sometimes we may need a state for is "the work is done, but we're holding this ticket until <launch date> to deploy". any thoughts on all of this? (maybe we can simply the statuses customizable via root.config.ts)

jeremydw:
- In Jira you can customize the statuses with each project (we could consider letting the statuses be customized too).
- We can have a higher-level state of an issue, e.g. "Open" "Closed" "Archived" separate from the status field

Erm, maybe the status field could have these by default?

- Backlog
- New
- In Progress
- In Review
- Done ("complete")
- Abandoned ("won't do")

stevenle:
trying to figure out the flow, especially in the kanban board.

```
new -> assign to <person> -> in progress -> <completes work> -> in review
```

from here i suppose github PRs have separate "approval" flag which is nice, so maybe we can add an `approvedBy: string[]` field. once it's approved and pending go-live, what do you think should happen? should it stay in "in review" but maybe we label it as "in review (approved)"?

also trying to see if there's a softer word for "abandoned" that won't trigger people.

stevenle:
decided to use the following statuses for now:

New
In Progress
In Review
Closed

went with github's convention for just having a single "closed" status (one of the llms made a good point to me that the last comment probably matters more than the status label itself)

will add support for custom statuses in a future release.

PR #1077, opened 2026-04-30
chore: improve a few edge cases related to new environment setup

jeremydw:
- fixes an issue with the env-init script when the prerequisites aren't satisfied
- replaces the "loading..." message with an error message when the loading fails outright

### DIFF
--- a/packages/root-cms/signin/signin.tsx
+++ b/packages/root-cms/signin/signin.tsx
@@ -287,7 +287,7 @@ const root = document.getElementById('root')!;
 function getStartupErrorMessage(err: any): string {
   const code = err?.code || '';
   if (code === 'auth/invalid-api-key') {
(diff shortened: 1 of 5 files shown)


packages/root-cms/signin/signin.tsx
```diff
@@ -282,9 +282,40 @@ function loginSuccessRedirect() {
...
-const auth = getAuth(app);
-window.firebase = {app, auth};
 const root = document.getElementById('root')!;
-root.innerHTML = '';
-render(<SignIn />, root);
+
+function getStartupErrorMessage(err: any): string {
+  const code = err?.code || '';
+  if (code === 'auth/invalid-api-key') {
+    return 'Firebase API key is invalid. Check the CMS firebaseConfig in root.config.ts.';
```
stevenle: can you make sure this only displays in dev? i wouldn't want this string leaking publicly.

stevenle (approved):

PR #1084, opened 2026-05-04
feat: add picker variant + presets to schema.oneOf()

### LINKED ISSUE #1068: Component / module picker variant of schema.oneOf()
**Description**

The schema.oneOf() field should support a more rich component picker, with screenshots and user-friendly descriptions. Optionally, there should be a "prefilled" values ("presets"?) that can be configured.

Old mockup from ~4 yrs ago:

Image

Although the screenshot shows a multi-column layout, I would prefer a single-column layout here since the multi-column hurts muscle memory when new options are added and cards shift positions.

stevenle:
## Summary
Introduces a new `ComponentPickerModal` component and adds support for schema presets, enabling a visual picker UI for `oneOf` fields as an alternative to the dropdown variant.

## Key Changes

- **New ComponentPickerModal component**: A searchable modal that displays component options as cards with thumbnails, titles, descriptions, and optional preset variants
  - Supports filtering by schema name, label, description, preset id, and preset label/description
  - Displays empty state when no options match the search
  - Integrates with Mantine's modal system via `useComponentPickerModal` hook

- **Schema preset support**: Added `SchemaPreset` interface and `presets` array to `Schema`
  - Presets include id, label, description, image, and prefill data
  - Schemas can now have an optional `image` field for thumbnail representation
  - Presets appear as separate cards in the picker alongside the schema's blank card

- **OneOfField picker variant**: Added `variant` option to `OneOfField` ('dropdown' or 'picker')
  - When `variant: 'picker'`, displays a button that opens the component picker modal
  - Supports clearing the selected type
  - Preset data is deep-merged over schema defaults to preserve nested object structure

- **Deep merge utility**: Implemented `deepMerge` function for recursively merging objects
  - Handles nested object merging while replacing arrays wholesale
  - Skips undefined source values but allows null to override
  - Non-mutating implementation

- **UI improvements**: Added styling for the picker button layout and modal cards with hover states

## Implementation Details

- The picker modal uses `useMemo` to efficiently filter options based on search query
- Preset data is applied via a new `applyType` function that deep-merges prefill data over schema defaults
- The component picker modal is registered with Mantine's modal system using a stable `MODAL_ID`
- Card rendering extracts subtitle (schema name) when displaying presets to provide context
- Comprehensive test coverage for preset functionality and deep merge behavior

## Screenshots

Screenshot 2026-05-04 at 7 41 50 AM

## References

Fixes #1068 

https://claude.ai/code/session_01JTWi28i8Eztd7bjQYvvLEg

### DIFF
--- a/docs/collections/Pages.schema.ts
+++ b/docs/collections/Pages.schema.ts
@@ -50,6 +50,7 @@ export default schema.collection({
           help: 'Compose the page by adding one or more modules.',
           of: schema.oneOf({
             types: schema.glob('/templates/*/*.schema.ts'),
+            variant: 'picker',
           }),
           preview: [
             'm{_index:02}: {description} (#{id})',
--- a/packages/root-cms/core/schema.ts
+++ b/packages/root-cms/core/schema.ts
@@ -358,9 +358,6 @@ export interface SchemaPreset {
    * Field values to prefill. Deep-merged into the schema's default value
    * (preset values override defaults). `_type` is set automatically. The
    * preset id is NOT persisted — presets are purely a prefill mechanism.
-   *
-   * For `array` fields, use the array-map shape
-   * `{_array: ['k1'], k1: {...}}` so the array UI renders correctly.
    */
   data?: Record<string, any>;
 }
@@ -425,6 +422,30 @@ export function defineSchema(schema: Schema): Schema {
 
 /** Defines the schema for a collection or reusable component. */
 export const define = defineSchema;
+
+/**
+ * Defines a preset for use within `schema.define({presets: [...]})`. The
+ * optional generic parameter `T` constrains the shape of the preset's `data`
+ * field. Pass an auto-generated fields type to enable type-safety:
+ *
+ * ```ts
+ * import type {HeroFields} from './generated/types.d.ts';
+ *
+ * schema.definePreset<HeroFields>({
+ *   id: 'big',
+ *   label: 'Big hero',
+ *   data: {title: 'Welcome'},
+ * });
+ * ```
+ */
+export function definePreset<T = any>(
+  preset: Omit<SchemaPreset, 'data'> & {data?: T}
+): SchemaPreset {
+  return preset as SchemaPreset;
+}
+
+/** Alias for {@link definePreset}. */
+export const preset = definePreset;
 
 export type Collection = SchemaWithTypes & {
   /**
--- a/packages/root-cms/core/schema.ts
+++ b/packages/root-cms/core/schema.ts
@@ -345,7 +345,7 @@ export type ObjectLikeField =
   | OneOfField
   | ReferenceField;
 
-export interface SchemaPreset {
+export interface SchemaPreset<T = Record<string, any>> {
   /** Unique identifier for the preset within the schema. */
   id: string;
   /** Display title for the preset. Defaults to the preset id. */
@@ -358,8 +358,11 @@ export interface SchemaPreset {
    * Field values to prefill. Deep-merged into the schema's default value
    * (preset values override defaults). `_type` is set automatically. The
    * preset id is NOT persisted — presets are purely a prefill mechanism.
+   *
+   * Pass a generated fields type as the generic parameter `T` to
+   * type-check the keys against the schema's fields.
    */
-  data?: Record<string, any>;
+  data?: T;
 }
 
 export interface Schema {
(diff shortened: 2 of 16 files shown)


packages/root-cms/core/schema.ts
```diff
@@ -338,13 +345,43 @@ export type ObjectLikeField =
...
+  /** Description shown below the title in the picker. */
+  description?: string;
+  /** Thumbnail image (URL or absolute /-path) for the picker. */
+  image?: string;
+  /**
+   * Field values to prefill. Deep-merged into the schema's default value
+   * (preset values override defaults). `_type` is set automatically. The
+   * preset id is NOT persisted — presets are purely a prefill mechanism.
+   */
+  data?: Record<string, any>;
```
jeremydw: Can we assign a stronger type than `any` here?

docs/collections/Pages.schema.ts
```diff
@@ -50,6 +50,7 @@ export default schema.collection({
           help: 'Compose the page by adding one or more modules.',
           of: schema.oneOf({
             types: schema.glob('/templates/*/*.schema.ts'),
+            variant: 'picker',
```
jeremydw, line 53: Also add presets for 1-2 templates in the `docs` package.

jeremydw:
Looking good! Two pieces of feedback:

1. Can we add a segmented layout picker in the top right, next to the search field? One icon for list (default) and one for grid. If grid is selected, it should show the results in a 3x3 grid. And it should remember the last selection in local storage.

2. Tags - we can do this in a followup but for big sites it'd be good to have tagging support for templates. The tags could show as chips at the top of the picker, below the search box, so when you click on one of the tags it filters the list automatically.

3. Make sure the search field is autofocused so you don't have to click on the field when the picker modal opens. And ideally down/up arrows work so you can navigate the whole thing quickly with the keyboard (same as the original "dropdown" ui).

Also regarding the `variant: 'picker'` do you think we should just make this automatic? If any of the templates have descriptions, images, presets, etc. then show the picker.

stevenle:
grid is done:

Screenshot 2026-05-04 at 12 26 34 PM

categories/tags we can do in a separate PR. i think it'll be nice to group certain "helpers" together (spacers, containers, etc.) vs. "page module templates" vs. "one-off custom modules". tags will be nice for filtering, e.g. "carousels".

need to think about the default, i worry users will find it frustrating that suddenly a possibly undesirable ux change happened and they're not sure how to revert. the popup is a nicer experience imo but some power users may prefer the quickness of the dropdown version. (maybe we make it a sitewide root.config.ts setting that allows users to define. or i can set up a per user setting in /cms/settings)

jeremydw:
Sg re: followup for tagging.

Some more thoughts:

1. What does "Remove component" do in the modal?
2. For the default; I was thinking we'd actually _leave_ the dropdown as-is, but add an _extra_ button next to it that, when clicked, opens the picker. (This button would only appear if we have the metadata). Basically, users can use the current dropdown OR the modal picker.

image

(For example, something like this would appear ^)

stevenle:
"remove component" clears the field value (setting to `{_type: ''}`. (e.g. if you have an oneOf() asset field, you may want to remove the asset altogether.)

the icon feels a little cluttered to me, i think we should just have one button here for selecting the type otherwise there's too many "choices" for the user to make.

jeremydw:
- Sounds good; we can give this a try. I think in most cases the second we have metadata added for a schema we'll want to switch over to use the modal picker.

- For "Remove component", hm that's a bit weird. I wonder if we should have a clear button outside the modal instead ...

stevenle:
claude had originally had the clear button next to the button like you suggested, but i thought it was 1) too busy for the ui, and 2) error prone - imagine clicking through a bunch of stuff and accidentally hitting the "clear" button, it would be a frustrating experience. i think having it nested deep down in the modal prevents accidentally clicking it.

jeremydw:
Got it - maybe if we rename the label to something like "Clear selection" that will help

jeremydw (approved):

jeremydw:
(Sorry, not sure why those two inline comments didn't send earlier. LGTM aside from that)

stevenle:
done. added a generic for the schema preset interface (users can use the generated type from root-cms.d.ts). claude added some nice presets for the 50x50 template (image + copy, copy + image).

PR #1096, opened 2026-05-07
ci: add docs-preview workflow for deploying to the preview channel

jeremydw:
- Runs a preview deployment when a project maintainer uses sa `/preview` slash command in a comment (https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/about-slash-commands, https://johnnyreilly.com/slash-command-your-deployment-with-github-actions)
- While the hosting channel automatically expires in 2d the underlying function (needed for SSR) doesn't get cleaned up automatically so we need to add a companion `docs-preview-cleanup.yaml` workflow that automatically deletes any of the `previewpr` deployments

### DIFF
--- a/.github/workflows/docs-preview-cleanup.yml
+++ b/.github/workflows/docs-preview-cleanup.yml
@@ -1,0 +1,85 @@
+name: Cleanup rootjs.dev Previews
+
+on:
+  schedule:
+    - cron: "17 * * * *"
+  workflow_dispatch:
+
+defaults:
+  run:
+    shell: bash
+
+env:
+  PREVIEW_MAX_AGE_HOURS: "48"
+
+jobs:
+  cleanup:
+    name: Delete Expired Preview Functions
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
+    steps:
+      - name: Authenticate to Google Cloud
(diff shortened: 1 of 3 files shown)


jeremydw:
FYI: The slash commands only work from `main` apparently so this won't work until merged, then I think we can try it on our active PRs (though they might need to be rebased first, not sure)

stevenle:
1) can you update the build have the local version updated to maybe the branch name so that it displays on the top left corner of the root cms ui? that'll make it clear to us that the preview build is working correctly.

2) i see an error in a different action because the two actions conflict / get out of sync, can you confirm this will fix it?
https://github.com/blinkk/rootjs/actions/runs/25474583638

3) is there a way to have this workflow not run automatically, but on demand instead? there are lots of times i do refactors that don't need to be staged and i don't wanna have this workflow run unnecessarily.

jeremydw:
1. Yup, good idea - just added some code that will rewrite the version number to include the pr number, e.g:

```
@blinkk/root       2.5.16-pr.<PR>
```

2. Modified the docs deployment to scope it to just the functions required for the docs deployment, so this should remove that error.

3. Yup, it already only runs when a project maintainer adds a comment to a PR that contains `/preview`. (It won't work until the workflow is on `main` though.)

PR #1099, opened 2026-05-07
ci: iterate on local bundling of root packages to fix pre-release deploys

jeremydw:


.github/workflows/docs-preview.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     if: >-
       github.event_name == 'workflow_dispatch' ||
-      (github.event.issue.pull_request && startsWith(github.event.comment.body,
+      (github.event.issue.pull_request && contains(github.event.comment.body,
```
stevenle, line 26: this might be a problem since the way i write i tend to say stuff like "the staging/preview link" which would trigger the build unintentionally.

can this be a little more deterministic?

jeremydw: changed it back to startswith

PR #1109, opened 2026-05-08
fix: improve performance of resizing the right-side panel

jeremydw:
- this would previously lock up on resource-intensive pages (lots of forms, complex preview, lots of checks)


stevenle:
@jeremydw i can merge this if you tested and there's a noticeable improvement on your end, but on my end these two feel about the same:

https://rootjs-dev--pr-1109-4k19as5d.web.app/cms/content/Pages/index
https://rootjs.dev/cms/content/Pages/index

(the prod one performs slightly better imo)

stevenle:
`alpha` branch actually introduces a second side panel, would you mind making this change in alpha? that way there wouldn't be a merge conflict.

PR #1137, opened 2026-05-18
chore: optimize JSX rendering

stevenle:
## Summary
This PR improves the performance of JSX rendering by introducing caching mechanisms and reducing unnecessary allocations in hot paths.

## Key Changes

- **Context caching**: Added version-based caching for the global context map (`buildGlobalContext`). The context is now only rebuilt when context values actually change, and reuses a shared empty object in the common case of no contexts.

- **Style key caching**: Introduced a `Map`-based cache for camelCase-to-kebab-case conversions in `toKebabCase`, avoiding repeated regex operations on the same style property names.

- **Block elements optimization**: Changed `blockSet` to reuse the module-level `DEFAULT_BLOCK_ELEMENTS` Set in pretty mode when no custom block elements are provided, eliminating unnecessary Set allocations.

- **Array rendering optimization**: Replaced `.map().join()` patterns with explicit loops when rendering arrays, reducing intermediate array allocations.

- **Props handling refactor**: Optimized the `jsx` function to avoid creating a new props object when `key` is not present in props, and to determine the final props before creating the VNode.

## Implementation Details

- The context caching uses a `ctxVersion` counter that increments on every `pushCtx`/`popCtx` call, allowing `buildGlobalContext` to quickly determine if a rebuild is needed.
- The `EMPTY_GLOBAL_CTX` shared object is safe to reuse since components only read from the context map.
- Style key caching uses a module-level `Map` that persists across renders, providing benefits for repeated style applications.
- The array rendering changes trade slightly more verbose code for better memory efficiency in the common case.

https://claude.ai/code/session_01XA3suqzHD9c6szRhZXgYQv

### DIFF
--- a/packages/root/src/jsx/jsx-render.ts
+++ b/packages/root/src/jsx/jsx-render.ts
@@ -162,6 +162,11 @@ const LT = '&lt;';
 const GT = '&gt;';
 const QUOT = '&quot;';
 
+// Shared empty object reused as the per-component context map when no contexts
+// have ever been pushed. Components only ever read from this map, so sharing is
+// safe and avoids a per-render allocation in the common case.
+const EMPTY_GLOBAL_CTX: Record<string, any> = {};
+
 /** Returns `true` when `value` is neither `null` nor `undefined`. */
 function isDef<T>(value: T | null | undefined): value is T {
   return value !== null && value !== undefined;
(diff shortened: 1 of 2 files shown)


stevenle:
```
Summary
Five targeted optimizations to @blinkk/root/jsx, all in jsx-render.ts / jsx-runtime.ts:

Cache buildGlobalContext — was rebuilt (with a fresh {}) per function-component render. Now a version counter on pushCtx/popCtx lets sibling renders reuse the same map, and a shared frozen-ish empty object is returned when no contexts have been pushed.
Reuse DEFAULT_BLOCK_ELEMENTS directly when no custom block elements are passed, and skip building blockSet entirely in minimal mode.
Memoize style-key camelCase→kebab-case conversion module-wide so the regex callback runs once per unique CSS property name.
for loops instead of .map().join('') in render and renderChildren — drops the intermediate string array and the per-call closure.
jsx() key handling — read props.key once, construct the VNode with the stripped props directly.
Result: ~19% faster pretty mode and ~23% faster minimal mode on a 500-component bench. All 79 tests pass; no new lint findings.
```

PR #1174, opened 2026-05-22
fix: jsx short circuit issue for 0x0 dimensions

jeremydw:


stevenle:
this fix doesn't look right. if the value were null it would short circuit with that logic already. and also i think this pr introduces an eslint error.

jeremydw:
updated to use `Number.isFinite`. re:lint - i saw that it reformatted four lines, but on my local that's how it wants to format those lines. :\ not sure what's going on there.

stevenle:
hmm have you `pnpm install`'ed lately? here's what i see when i pull this branch:

```
> eslint .


rootjs/packages/root-cms/ui/components/DocEditor/fields/FileField.tsx
  1080:9   error  Insert `··`                                                                 prettier/prettier
  1081:1   error  Insert `··`                                                                 prettier/prettier
  1297:18  error  Replace `⏎······'--canvas-aspect-ratio'` with `'--canvas-aspect-ratio']·=`  prettier/prettier
  1299:5   error  Replace `]·=` with `·`                                                      prettier/prettier

✖ 4 problems (4 errors, 0 warnings)
  4 errors and 0 warnings potentially fixable with the `--fix` option.
```

i'm also a little curious why the `?.width && ?.height` check is failing here. is the db corrupted with bad values?

jeremydw:
- Restarted my IDE (and machine) which fixed the lint issue.
- Re `?.width && ?.height` I think it's because of how JSX treats the number 0. The issue wasn't so much that the integrity of the DB was incorrect but the file was written with zero width and height. So if the field values are literally 0 it will emit the 0!

image

stevenle:
i see what you're saying but isn't a value of 0 in the width/height fields an indication of bad data in the db? in what cases would we display `0x0` here? this PR would allow for that, which seems incorrect to me. so maybe instead of the isFinite() check we can do `Boolean(width && height) && (...)`

jeremydw:
done

stevenle (approved):

PR #1186, opened 2026-05-28
feat: switch field metadata to `{translate: false}` and add `{alt: false}` toggle

stevenle:
## Summary
This PR introduces a metadata-driven approach for controlling alt text and translation settings on a per-field basis. It replaces the `disableTranslations` metadata key with a more semantic `translate: false` convention and adds support for disabling alt text on individual image fields through metadata.

## Key Changes

- **Alt text metadata control**: Added `altDisabledByMetadata` and `setAltDisabledByMetadata` to `FileFieldContextValue` and `FileFieldInternalProps`, allowing users to disable alt text input on a per-instance basis via a menu toggle
- **Metadata key generation**: Introduced `getMetadataKey()` utility function that converts field keys (e.g., `fields.hero`) to metadata keys (e.g., `fields.@hero`)
- **Translation metadata standardization**: Replaced `disableTranslations` with `translate: false` across all translation-related code for consistency and clarity
- **Extract utilities enhancement**: Updated `extractField()` and `extractFieldWithMetadata()` to accept and respect metadata, specifically checking `metadata?.alt !== false` before extracting image alt text
- **Translation checks**: Updated `checks-translations.ts` to use the new `translate: false` metadata convention
- **UI updates**: Added a "Disable alt text" menu item in the FileField preview that appears when editing is allowed and the setter is provided

## Implementation Details

- The metadata system uses a naming convention where field metadata is stored with an `@` prefix (e.g., `@fieldName`)
- Alt text can be disabled per-field by setting `metadata.alt = false`, which prevents both the UI input and translation extraction
- The translation disable/enable logic now uses `metadata.translate === false` instead of `metadata.disableTranslations`
- Metadata is only persisted when it contains non-default values to avoid unnecessary data storage
- The alt text disable toggle is only shown for image fields when the field schema allows alt text (`field.alt !== false`)

https://claude.ai/code/session_01Fa7Z7JbaB3U7ouQKduN3nt

### DIFF
--- a/packages/root-cms/core/checks-translations.ts
+++ b/packages/root-cms/core/checks-translations.ts
@@ -35,7 +35,7 @@ function extractTranslatableStrings(
 
       const metadataKey = `@${field.id}`;
       const metadata = fieldData[metadataKey];
-      if (metadata?.translate === false) {
+      if (metadata?.translate === false || metadata?.disableTranslations) {
         continue;
       }
 
--- a/packages/root-cms/core/checks-translations.ts
+++ b/packages/root-cms/core/checks-translations.ts
@@ -33,18 +33,21 @@ function extractTranslatableStrings(
         continue;
       }
 
-      // Check for "do not translate" metadata.
       const metadataKey = `@${field.id}`;
       const metadata = fieldData[metadataKey];
-      if (metadata?.disableTranslations) {
+      if (metadata?.translate === false) {
         continue;
       }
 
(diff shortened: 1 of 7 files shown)


stevenle:
@jeremydw fyi two changes here:

1) ability to disable alt text (via context menu, adds `{alt: false}` to the metadata)
2) rename `{disableTranslations: true}` to `{translate: false}` (shorter, cleaner)

PR #1190, opened 2026-05-31
fix: update the reload url used by the iframe preview

jeremydw:
This PR slightly tweaks the URLs used when the iframe preview refreshes after content changes.

- Previously, it would use the original URL. This would replace any client-side URL changes made by the child page (e.g. if the user uses a feature flag picker to change the flags on a page, the flags in the URL parameters would be wiped out on save.)
- Now, it reuses the iframe's existing URL (and for good measure, just forces `preview=true`).

### DIFF
--- a/packages/root-cms/ui/pages/DocumentPage/DocumentPage.tsx
+++ b/packages/root-cms/ui/pages/DocumentPage/DocumentPage.tsx
@@ -729,17 +729,12 @@ DocumentPage.Preview = (props: PreviewProps) => {
         !iframeWindow.location.href.startsWith('about:blank')
       ) {
         const currentUrl = iframeWindow.location;
-        // Strip `preview=true` from the displayed URL so the URL bar mirrors
-        // the public/prod URL. This avoids editors accidentally copying a
-        // preview-only URL into places like social media.
-        const displaySearchParams = new URLSearchParams(currentUrl.search);
-        displaySearchParams.delete('preview');
-        const displaySearch = displaySearchParams.toString();
-        setIframeUrl(
-          `${domain}${currentUrl.pathname}${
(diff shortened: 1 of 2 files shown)


packages/root-cms/ui/pages/DocumentPage/DocumentPage.tsx
```diff
@@ -659,10 +681,11 @@ DocumentPage.Preview = (props: PreviewProps) => {
       return;
     }
     const iframe = iframeRef.current!;
+    const nextUrl = getReloadUrl(iframe, localizedPreviewUrl);
```
stevenle: i think it's fine to preserve the query params / hash, but we should make sure that it's rendering either the localized url or the default url for the page. it doesn't make sense for us to reload and preserve the url for an iframe where a user has navigated to an entirely new site for example.

stevenle:
two issues i noticed from the staging link:
https://rootjs-dev--pr-1190-mfvupyzu.web.app/cms/content/Sandbox/test-5

- the ?preview=true part should be removed from the url bar (the url bar should mimic the live prod url, having that preview=true part can cause issues with content editors copying/pasting the prod url to mediums like social media)
- the intl urls aren't being preserved (e.g. toggle a different locale and then make an edit)

jeremydw:
Should be fixed now - but I'm wondering if we should suppress the mutated query string from the visible URL bar in the iframe preview? (Yet, include it in the pop out URL, of course). LMK what you prefer.

stevenle:
yeah let's exclude it and keep it as a clean url bar for the purposes of keeping just the preview pane url bar minimal.

jeremydw:
Just gave it a test - looks good now

stevenle (approved):

PR #1203, opened 2026-06-08
fix: resolve pod assets in asset map

### LINKED ISSUE #1201: Pods: SSR render logs "could not find build asset" for pod route HTML pages
## Summary

When a pod contributes an HTML-rendering route (via `routesDir`), every SSR
render of that route logs:

```
could not find build asset: pod/<pod-name>/<relPath>
```

The page still renders correctly, but the log line appears on every request.

## Version

`@blinkk/root` 3.0.5 (pods feature)

## Cause

`collectPods()` registers each pod route with a virtual `src` of the form
`pod/${pod.name}/${route.relPath}` (route-manifest generation in the pod
collector).

At render time, the renderer walks the route's dependency graph via
`BuildAssetMap.get(src)` to collect client JS/CSS to inject. The `BuildAssetMap`
is keyed from the Vite client manifest, which uses the route's **real** file
path — not the `pod/<name>/...` virtual prefix. The lookup therefore misses and
falls through to:

```js
console.log(`could not find build asset: ${src}`);
return null;
```

(`src/render/asset-map/build-asset-map.ts`, `BuildAssetMap.get`).

## Expected

Pod route `src`s should resolve to their built client asset — e.g. map the
`pod/<name>/...` virtual src back to the real file path before the asset-map
lookup (similar to how `realPathRelativeTo` is already used to handle symlinked
workspace packages). Routes that legitimately have no client bundle should
resolve cleanly without logging an error.

## Repro

1. Create a pod with a `routesDir` containing an HTML page route (a default
   exported component plus a `handle`).
2. Register the pod via a plugin, then run a production build and serve.
3. Request the pod route → observe `could not find build asset: pod/<name>/...`
   logged on each render.

## Impact

Cosmetic (log spam) for SSR-only pod routes. It could also mask genuinely
missing client assets for pod routes that *do* ship a client bundle, since both
cases produce the same log line.

jeremydw:
Fixes #1201.

## Problem

When a pod contributes an HTML-rendering route (via `routesDir`), every SSR
render of that route logs:

```
could not find build asset: pod/<pod-name>/<relPath>
```

The page still renders, but the route's CSS deps aren't collected and the log
line appears on every request.

## Cause

Pod routes are registered in the router with a **virtual** src,
`pod/${pod.name}/${route.relPath}` (`pods-vite-plugin.ts`). At render time the
renderer calls `assetMap.get(route.src)` with that virtual src
(`render/render.tsx`). But `BuildAssetMap` is keyed from the Vite client
manifest, which keys the route by its **real** file path — so the lookup misses
and falls through to `console.log("could not find build asset: ...")`. The
existing `realPathRelativeTo` fallback can't help because `pod/<name>/...` is a
synthetic prefix, not a real path.

## Fix

`BuildAssetMap.fromViteManifest()` now accepts the list of pod routes and, after
populating the map from the manifest, registers an **alias** for each pod route:
the virtual `pod/<name>/...` src points at the asset built from the route's real
file path. This mirrors how the same function already double-registers element
source files under both their `relPath` and resolved real path.

Like user `routes/` entries, the alias has no asset URL of its own (a pod
route's client JS is injected via elements/bundles, not the route chunk), so its
`assetUrl` is set to `''`. The alias is written into `dist/.root/manifest.json`,
so it also applies to the prod SSR server (which rebuilds the map via
`fromRootManifest`).

Pod routes that ship no client asset (e.g. pure JSON API routes) are simply not
aliased; those never reach this lookup since they don't render HTML.

## Testing

- Added `test/build-asset-map.test.ts` covering both the alias-hit case (CSS
  deps resolved, no asset URL) and the no-asset case.
- `pnpm test` (162 passing), `tsc --noEmit`, eslint, and prettier all pass.

### DIFF
--- a/.changeset/pod-route-asset-alias.md
+++ b/.changeset/pod-route-asset-alias.md
@@ -2,4 +2,4 @@
 '@blinkk/root': patch
 ---
 
-Fix pod route asset resolution so SSR no longer logs "could not find build asset" for pod-contributed HTML routes, and the route's CSS deps are collected correctly.
+fix: resolve pod assets in asset map
--- a/.changeset/pod-route-asset-alias.md
+++ b/.changeset/pod-route-asset-alias.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root': patch
+---
+
+Fix pods: resolve pod route assets at render time so SSR no longer logs "could not find build asset: pod/<name>/..." for pod route HTML pages. The build now aliases each pod route's virtual `pod/<name>/...` src to the asset built from the route's real file path, so the route's CSS deps are collected correctly.
--- a/.changeset/pod-route-asset-alias.md
+++ b/.changeset/pod-route-asset-alias.md
@@ -2,4 +2,4 @@
 '@blinkk/root': patch
 ---
 
-Fix pods: resolve pod route assets at render time so SSR no longer logs "could not find build asset: pod/<name>/..." for pod route HTML pages. The build now aliases each pod route's virtual `pod/<name>/...` src to the asset built from the route's real file path, so the route's CSS deps are collected correctly.
+Fix pod route asset resolution so SSR no longer logs "could not find build asset" for pod-contributed HTML routes, and the route's CSS deps are collected correctly.
--- a/packages/root/src/cli/build.ts
+++ b/packages/root/src/cli/build.ts
@@ -295,10 +295,17 @@ export async function build(rootProjectDir?: string, options?: BuildOptions) {
 
   // Use the output of the client build to generate an asset map, which is used
   // by the renderer for automatically injecting dependencies for a page.
+  const podRoutes = pods.flatMap((pod) =>
+    pod.routeFiles.map((route) => ({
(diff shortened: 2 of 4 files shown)


stevenle:
@jeremydw can you clean up the PR title and changelog message and address copilot's comments? i'll merge after that.

jeremydw:
Thanks @stevenle — done in ee2971b:

- **Title + changelog** cleaned up (title is now "Resolve pod route assets in build asset map"; changeset trimmed to a single sentence).
- **Copilot #1/#2 (test):** imported `vi` and now spy on `console.log` (with `mockRestore`) in the no-asset test, so there's no log noise — and it asserts the "could not find build asset" message fires on that path.
- **Copilot #3 (separators):** normalized the `realPathRelativeTo()` candidate key to forward slashes so aliasing matches the manifest keys on Windows/symlinked setups.

`pnpm test` (162 passing), tsc, eslint, and prettier all green.

stevenle (approved):

PR #1205, opened 2026-06-08
feat: add `excludeLocalesFromTranslations` config to exclude locales from translation import/export

stevenle:
## Summary
This PR adds a new `excludeLocalesFromTranslations` configuration option to the CMS plugin that allows excluding specific locales from CSV import and export operations in the Localization modal. This is useful for excluding pseudo-locales or test locales (e.g., `ALL_es`, `ALL_fr`) from CSV workflows.

## Key Changes
- **New configuration option**: Added `excludeLocalesFromTranslations` to `CMSPluginOptions` in `plugin.ts` with support for wildcard patterns (`*` for any characters, `?` for single character)
- **Utility function**: Implemented `isCsvLocaleExcluded()` in `l10n.ts` that checks if a locale matches any exclude patterns with case-insensitive matching
- **Helper function**: Added `wildcardToRegExp()` to convert wildcard patterns to RegExp, properly escaping regex special characters while supporting `*` and `?` wildcards
- **CSV export filtering**: Modified `formatCsvData()` to exclude matching locales from the exported CSV headers and data
- **CSV import filtering**: Added logic to filter out excluded locale columns during CSV import so they aren't processed
- **Type definitions**: Updated `window.__ROOT_CTX` type in `ui.tsx` to include the new `csv` configuration
- **App initialization**: Updated `renderApp()` in `app.tsx` to pass the `csv.excludeLocales` config to the UI context
- **Comprehensive tests**: Added full test suite for `isCsvLocaleExcluded()` covering wildcard patterns, case-insensitivity, exact matches, multiple patterns, and literal regex character handling

## Implementation Details
- Patterns are matched case-insensitively against locale IDs
- Regex special characters in patterns are treated literally (e.g., `en.US` matches exactly, not as a regex)
- Both export and import operations respect the exclude list to maintain consistency
- The feature is backward compatible—if no patterns are configured, all locales are included

https://claude.ai/code/session_01MVfE3rU7rirVnLpiRehC8Z

### DIFF
--- a/.changeset/strong-locales-hide.md
+++ b/.changeset/strong-locales-hide.md
@@ -2,4 +2,4 @@
 '@blinkk/root-cms': patch
 ---
 
-feat: add `csv.excludeLocales` config to exclude locales from CSV import/export
+feat: add `excludeLocalesFromTranslations` config to exclude locales from translation import/export
--- a/.changeset/strong-locales-hide.md
+++ b/.changeset/strong-locales-hide.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: add `csv.excludeLocales` config to exclude locales from CSV import/export
--- a/packages/root-cms/core/app.tsx
+++ b/packages/root-cms/core/app.tsx
@@ -146,9 +146,8 @@ export async function renderApp(
       hasExport: isFunction(t.onExport),
     })),
     defaultOneOfVariant: cmsConfig.defaultOneOfVariant || 'dropdown',
-    csv: {
-      excludeLocales: cmsConfig.csv?.excludeLocales || [],
-    },
+    excludeLocalesFromTranslations:
+      cmsConfig.excludeLocalesFromTranslations || [],
   };
   const projectName = cmsConfig.name || cmsConfig.id || '';
   const title = getCmsTitle(projectName, cmsConfig.minimalBranding);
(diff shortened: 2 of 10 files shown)


stevenle:
@jeremydw this config is for excluding locales from being "exported" to sheets, mainly for aliased locales like "ALL_<country>". the config name was created by ai, i couldn't think of a better name, if you have suggestions lmk.

stevenle:
note: config name changed to `excludeLocalesFromTranslations` for more explicitness.

jeremydw:
LGTM - although I wonder if we should adopt the term "skip" instead of "exclude"? The idea is that the locales are still processed/generated/translations are stored as usual (right?) they're just skipped from the import/export workflows?

jeremydw (approved):

stevenle:
i think "exclude" feels more correct to me in the context of "excludeLocalesFromTranslations". 

there's some more work i need to do to get the "translations identifiers" distinguished from "locale identifiers", but basically:

- a site may define "locales" as `"<lang>_<country>"` (with fallbacks as `"ALL_<country>"` and `"<lang>_ALL"`)
- we want to be able to alias or map a "locale identifier" to its "translation identifier" equivalent, e.g. locale=fr_ca -> translator_lang=fr-CA, and locale=ALL_de -> translator_lang=de
- until we have the translator langs "mapping" i think "excludeLocalesFromTranslations" works fine for ignoring the fallback locales from the translator pipeline (this config may be needed even when we have translator language id mappings)

PR #1220, opened 2026-06-10
feat: parallel SSG rendering with `root build --threads`

jeremydw:
## Summary

`root build` currently renders all SSG pages on a single CPU core. The `-c/--concurrency` flag overlaps async I/O (e.g. CMS fetches in `getStaticProps()`), but JSX rendering itself is synchronous CPU work, so large sites (hundreds of pages across many locales) bottleneck on one core while the rest of the machine sits idle — on a typical multi-core CI machine, the render phase of a large site can take many minutes single-threaded.

This PR adds an opt-in `--threads [num]` flag that renders pages using a pool of `worker_threads`:

```sh
# Render pages using 8 worker threads.
root build --threads 8

# Pick a worker count automatically based on cpu cores and page count.
root build --threads
root build --threads auto
```

In auto mode, one core is reserved for the main thread and each worker must have enough pages to justify its fixed startup cost (loading the server bundle and manifests); small builds automatically stay in-process, since a single worker is never faster than rendering in-process.

## How it works

- The main thread builds the server/client bundles, computes the sitemap, and schedules pages; workers render them.
- Each worker bootstraps the same way the prod server (`root start`) does: it loads `dist/server/render.js`, `dist/root.config.js`, and the asset manifest / element graph from `dist/.root/`, then renders pages independently.
- Routes can't be structured-cloned across threads, so each task is identified by `{urlPath, params, routeSrc, locale}`. Workers resolve the route from a map keyed by `src + locale` built by walking the router (URL matching alone can't resolve extension routes like `[slug].json.ts`, so this avoids relying on it; URL matching is kept only as a fallback).
- The single-threaded code path is unchanged in behavior — the per-page logic was extracted to a shared `buildPage()` used by both paths, so output logging, sitemap.xml generation, `--filter`, `excludeDefaultLocaleFromIntlPaths`, and error reporting behave identically.
- `-c/--concurrency` is distributed across workers (e.g. `-c 30 --threads 6` gives each worker an async concurrency of 5), so per-page I/O still overlaps within each thread.
- Page errors are reported with the same detail as before (route src, params, stack) and fail the build; worker crashes reject all in-flight pages.

## Results

On a synthetic project (8 locales, 360 pages, ~3.5 MB rendered HTML per page), an M-series laptop:

| Configuration | Wall time | CPU time |
|---|---|---|
| `root build -c 30` | 7.9s | 8.0s (1 core saturated) |
| `root build -c 30 --threads 4` | 3.2s | 11.1s |
| `root build -c 30 --threads 8` | 3.5s | 14.9s |

The synthetic benchmark becomes disk-I/O bound at high thread counts; real-world sites with heavier component trees and more pages should scale further on high-core CI machines. Real-world builds also spend significant time in `getStaticProps()` data fetching, where the per-worker async concurrency continues to apply.

Output is byte-for-byte identical to single-threaded builds (verified by the included test, which exercises multiple locales, `getStaticPaths()`, `getStaticContent()` file routes, `404.tsx`, and sitemap.xml generation).

## Notes

- The flag is fully opt-in; builds without `--threads` use the existing code path.
- Workers share the same `dist/` artifacts and cwd, so filesystem-based caches that projects populate in `preBuild` hooks are shared across workers.
- `preBuild`/`postBuild` plugin hooks still run once, on the main thread.
- Worker memory: each thread loads its own copy of the server bundle; memory scales roughly linearly with thread count. The default (`--threads` with no value) reserves one core for the main thread.

## Test plan

- [x] New test: `packages/root/test/build-threads.test.ts` builds a fixture with and without `--threads 2` and asserts byte-identical output (multi-locale pages, `getStaticPaths`, `getStaticContent` file routes, 404, sitemap.xml).
- [x] Full `@blinkk/root` test suite passes (31 files / 163 tests).
- [x] `examples/starter` builds with `--threads 2` with output identical to the single-threaded build.
- [x] Lint/prettier clean; `pnpm build` (esbuild + tsc) passes.

### DIFF
--- a/.changeset/build-threads.md
+++ b/.changeset/build-threads.md
@@ -1,38 +1,5 @@
 ---
-'@blinkk/root': minor
+'@blinkk/root': patch
 ---
 
 feat: parallel SSG rendering with `root build --threads`
-
-`root build` now accepts a `--threads [num]` flag that renders SSG pages
-using N worker threads. Page rendering was previously limited to a single
-CPU core — the `-c/--concurrency` flag overlaps async I/O (e.g. CMS fetches
-in `getStaticProps()`) but JSX rendering itself is synchronous CPU work, so
-builds for large sites (hundreds of pages across many locales) were
-bottlenecked on one core.
-
-Each worker thread loads the site's built server bundle
-(`dist/server/render.js`) plus the asset manifest and renders pages
-independently, mirroring how the prod server bootstraps. The main thread
-computes the sitemap, schedules pages across workers, and aggregates
-results (output logging and sitemap.xml generation are unchanged).
-
-Usage:
-
-```sh
-# Render pages using 8 worker threads.
-root build --threads 8
-
-# Pick a worker count automatically based on cpu cores and page count.
-root build --threads
-root build --threads auto
-```
-
-In auto mode, one core is reserved for the main thread and each worker must
-have enough pages to justify its startup cost; small builds automatically
-stay in-process.
-
-The flag is opt-in; builds without `--threads` behave exactly as before.
-`-c/--concurrency` is distributed across workers (e.g. `-c 30 --threads 6`
-gives each worker an async concurrency of 5). Output is byte-for-byte
-identical to single-threaded builds.
--- a/.changeset/build-threads.md
+++ b/.changeset/build-threads.md
@@ -1,0 +1,33 @@
+---
+'@blinkk/root': minor
+---
+
+feat: parallel SSG rendering with `root build --threads`
+
+`root build` now accepts a `--threads [num]` flag that renders SSG pages
+using N worker threads. Page rendering was previously limited to a single
+CPU core — the `-c/--concurrency` flag overlaps async I/O (e.g. CMS fetches
+in `getStaticProps()`) but JSX rendering itself is synchronous CPU work, so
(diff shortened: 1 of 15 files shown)


stevenle:
if the perf boost has no downsides i'd like to make `--threads=auto` the default behavior but i need to do some due diligence to ensure there are no oddities with the change. iirc threads don't share memory by default so i don't know how that might affect sites that use caching for their ssg builds, will need to look closer at the code to see how things like that are handled.

jeremydw:
> if the perf boost has no downsides i'd like to make `--threads=auto` the default behavior but i need to do some due diligence to ensure there are no oddities with the change. iirc threads don't share memory by default so i don't know how that might affect sites that use caching for their ssg builds, will need to look closer at the code to see how things like that are handled.

Yeah; I was thinking we'd make it opt-in to start and then switch it (soon) after doing some validation and making sure all the various CI flows (GCB, Gitlab Pipelines, GitHub Actions, etc.) work as expected (and faster!)

stevenle:
ok yeah per chat i think this'll work, although users that rely on shared memory for speedups in their ssg builds will need to be aware of the change if we were to make --threads=auto the default.

per your suggestion i'll merge this in as is and we can look into the auto parallelism in the future.

PR #1221, opened 2026-06-10
feat: brotli compression for prod/preview server responses

jeremydw:
## Summary

The prod (`root start`) and preview (`root preview`) servers compress responses using `compression@1.7.4`, which predates brotli support. Clients that advertise only `Accept-Encoding: br` receive **entirely uncompressed responses** — observed in production as a ~200 kB HTML page going over the wire raw to brotli-preferring clients. Clients that accept gzip get gzip, but brotli typically compresses HTML 20-30% smaller.

This PR:

1. **Upgrades `compression` to 1.8.1** (adds brotli encoding negotiation).
2. **Adds a shared `compressionMiddleware()`** (exported from `@blinkk/root/middleware`) used by both `root start` and `root preview`, configured with **brotli quality 4** — the standard setting for on-the-fly dynamic compression. The brotli library default (q11) is deliberately avoided: it is ~10-50x more CPU-expensive and only suitable for compress-once static assets.

## Measured results

End-to-end against `root start` serving a 3.8 MB rendered HTML page (localhost, so timings are dominated by compression CPU — a conservative comparison):

| Encoding | Wire size | Total response time |
|---|---|---|
| brotli q4 | 27.6 kB (**-63% vs gzip**) | ~6.4 ms (**~2x faster than gzip**) |
| gzip (level 6 default) | 74.0 kB | ~11.8 ms |
| identity | 3,790 kB | — |

Brotli output verified byte-identical to the uncompressed response after decompression. Clients without `br` support fall back to gzip exactly as before; responses under the existing 1 kB threshold remain uncompressed.

## Notes

- `Vary: Accept-Encoding` continues to be set by the middleware, so CDN/proxy caches key correctly per encoding.
- Decompression speed on the client is roughly constant regardless of encode quality, so users only see the smaller payload.
- No API changes for existing projects; the middleware swap is internal to the servers. Projects that want different brotli settings can layer their own middleware via `configureServer` as before.

## Test plan

- [x] Full `@blinkk/root` test suite passes (30 files / 162 tests).
- [x] E2E verification with `root start`: `br`, `gzip`, and `identity` requests return correct `Content-Encoding`, correct sizes, and byte-identical content after decompression.
- [x] Small responses (<1 kB) correctly skip compression.
- [x] Lint/prettier clean; `pnpm build` (esbuild + tsc) passes.

### DIFF
--- a/.changeset/brotli-compression.md
+++ b/.changeset/brotli-compression.md
@@ -3,17 +3,3 @@
 ---
 
 feat: brotli compression for prod/preview server responses
-
-Upgrades `compression` to 1.8.1 and configures the prod (`root start`) and
-preview (`root preview`) servers to serve brotli-encoded responses when the
-client sends `Accept-Encoding: br`, falling back to gzip as before.
-
-Previously, `compression@1.7.4` did not understand brotli at all — clients
-that advertised only `br` received entirely uncompressed responses.
-
-Brotli quality is set to 4, the standard setting for on-the-fly compression
-of dynamic responses. In an end-to-end test against `root start` with a
-3.8 MB rendered page, brotli q4 produced a 63% smaller payload than gzip
-(27.6 kB vs 74 kB) at lower CPU cost (~6ms vs ~12ms total response time).
-Note that brotli's library default (q11) is intentionally avoided — it is
-~10-50x more expensive and only appropriate for compress-once static assets.
--- a/.changeset/brotli-compression.md
+++ b/.changeset/brotli-compression.md
@@ -1,0 +1,22 @@
+---
+'@blinkk/root': patch
+---
+
+feat: brotli compression for prod/preview server responses
+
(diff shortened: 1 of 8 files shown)


stevenle:
can you remove the getTranslations() memoization from this PR? we don't want to create memory leaks with translations - a large site with large number of locales and large number of source strings would have problems with this approach.

stevenle:
note that `import.meta.glob()` is resolved at compile time so i don't think that memoization helps much.

jeremydw:
Done — removed the `getTranslations()` memoization in f2ca5d1b (file reverted to main). Good call: since `import.meta.glob(..., {eager: true})` is resolved at compile time, the per-call work was just the map-building loop, and retaining all locales' strings for the process lifetime is a bad trade for large multi-locale sites. The PR is now scoped to the compression change only.

PR #1229, opened 2026-06-12
feat: add `i18n.translationLanguages` config to map root locales to translation languages

### LINKED ISSUE #1216: Distinguish between "locales" and "translation languages"
**Description**

Translations services often use different identifiers than locales do. For the purposes of the site, a "locale" is considered a language + country combination used for routing purposes, whereas translations typically only handle the "language" portion (with optional variants, e.g. "fr-FR" vs. "fr-CA").

The root.config.ts needs to be able to convert between the two when importing/exporting translations from various services. The configuration should also support some fallback system for when a translation doesn't exist, e.g. en-CA -> en-GB -> en, es-MX -> es-419 -> en.

stevenle:
## Summary

This PR introduces a new `i18n.translationLanguages` configuration option that allows mapping root locales to translation language identifiers. This enables multiple root locales to share the same translations (e.g., `es-419_mx` and `es-419_co` both using `es-419` translations).

## Key Changes

- **New shared utilities** (`packages/root-cms/shared/translation-languages.ts`):
  - `getTranslationLanguage()` - converts a root locale to its translation language
  - `toTranslationLanguages()` - dedupes locales that share a translation language
  - `getLocalesForTranslationLanguage()` - expands a translation language to all matching root locales
  - `getTranslationForLanguage()` - reads translations from a map, checking all locales in a language group

- **UI layer wrappers** (`packages/root-cms/ui/utils/l10n.ts`):
  - Added wrapper functions that read the i18n config from `window.__ROOT_CTX`
  - Updated `updateTranslationByHash()` and `batchSaveTranslations()` to use `getLocalesForTranslationLanguage()` for expanding language columns to root locales

- **CMS components updated** to operate on translation languages instead of root locales:
  - `LocalizationModal` - shows translation language options, uses `getTranslationForLanguage()` to read values
  - `EditTranslationsModal` - re-keys translation rows from root locales to translation languages
  - `TranslationsTable` - displays translation language columns
  - `DocTranslationsPage` - filters and displays translation languages
  - `TranslationsEditPage` - edits keyed by translation language

- **Import/export logic** (`packages/root-cms/ui/utils/doc.ts`):
  - Updated `cmsDocImportTranslations()` to handle both root locale and translation language columns
  - Language columns fan out to all matching root locales, with direct locale columns taking precedence

- **Translation service context** (`packages/root-cms/core/translations.ts`):
  - Added `translationLanguages` field to `TranslationServiceContext`
  - Updated documentation to clarify that `TranslationRow.translations` keys are translation languages

- **Config schema** (`packages/root/src/core/config.ts`):
  - Added `translationLanguages?: Record<string, string>` to `RootI18nConfig`

- **Translation service plugins** updated to use `ctx.translationLanguages` when available (Crowdin, DeepL)

- **Test mocks** updated to support the new translation language utilities

## Implementation Details

- All matching is case-insensitive for locale/language identifiers
- When no mapping is configured, locale IDs are used as translation languages
- The conversion applies consistently across CSV/Google Sheets import/export, CMS UI, and translation service integrations
- Locales excluded from translations (e.g., `ALL_*` patterns) are properly filtered during import/export

Fixes #1216

https://claude.ai/code/session_01Wd2wVeaYVy66fF91KKgBeC

### DIFF
--- a/packages/root-cms/ui/utils/l10n.ts
+++ b/packages/root-cms/ui/utils/l10n.ts
@@ -9,6 +9,12 @@ import {
   where,
   writeBatch,
 } from 'firebase/firestore';
+import {
+  getLocalesForTranslationLanguage as sharedGetLocalesForTranslationLanguage,
+  getTranslationForLanguage as sharedGetTranslationForLanguage,
+  getTranslationLanguage as sharedGetTranslationLanguage,
+  toTranslationLanguages as sharedToTranslationLanguages,
+} from '../../shared/translation-languages.js';
 import {logAction} from './actions.js';
 import {TIME_UNITS} from './time.js';
 
@@ -94,8 +100,9 @@ export async function updateTranslationByHash(
       updates.tags = translations.tags;
       continue;
     }
-    const locale = normalizeLocale(key);
-    if (locale) {
+    // A key may be a root locale or a translation language shared by multiple
+    // root locales (e.g. `es-419` covering `es-419_mx` and `es-419_co`).
+    for (const locale of getLocalesForTranslationLanguage(key)) {
       updates[locale] = translations[key];
     }
   }
@@ -208,8 +215,7 @@ function wildcardToRegExp(pattern: string): RegExp {
  * Patterns support wildcards, e.g. `ALL_*`.
  */
 export function isLocaleExcludedFromTranslations(locale: string): boolean {
-  const patterns =
-    window.__ROOT_CTX.excludeLocalesFromTranslations || [];
+  const patterns = window.__ROOT_CTX.excludeLocalesFromTranslations || [];
   return patterns.some((pattern) => wildcardToRegExp(pattern).test(locale));
 }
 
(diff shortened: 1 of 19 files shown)


stevenle:
@jeremydw lmk if any feedback here.

packages/root-cms/ui/utils/l10n.ts
```diff
@@ -225,6 +231,49 @@ export function normalizeLocale(locale: string) {
...
+}
+
+/**
+ * Reads the translation for a translation language (or root locale) from a
+ * translations map keyed by root locale, checking all root locales that
+ * share the language and returning the first non-empty value.
+ */
+export function getTranslationForLanguage(
+  translations: Record<string, unknown>,
+  lang: string
```
jeremydw: @stevenle I wonder if we should assign type aliases; e.g. 

```
export type RootLocale = string;
export type RootLanguage = string;
```

Or maybe using generics and allowing sites to pass in the site's string literals as a bounded list of locales. One thing I really struggle with is remembering which string is which and it might help to avoid using strings as the types and instead use something that gives a stronger signal.

Even if it's not a generic it would be useful to have the code completion/comments that describe the difference between RootLocale and RootLanguage

stevenle: the problem is that root needs to remain agnostic of what a "locale" is and how it's formatted, if we provide examples here it signals that it is the de-facto / correct value when it might be configured differently for how the site serves.

firebase hosting locales for example is different from how another site might prefer handle it. some sites do fancy things like `ALL_<country>` fallbacks whereas some sites use `<lang>` as the locale identifiers.

jeremydw: Ah yeah, I guess casting the identifiers to/from Firestore queries would be more trouble than its worth. I asked AI and it sort of agreed that if we just do type aliases, it would work for both of our cases (documentation as well as just keeping the type as `string`.

```
/**
 * A site-defined locale identifier, as configured in `i18n.locales`.
 * Root is agnostic to its format.
 */
export type RootLocale = string;

/**
 * The language identifier used by translation systems (CSV/Sheets columns,
 * Crowdin/DeepL, CMS translations pages), mapped via
 * `i18n.translationLanguages`. Defaults to the root locale id.
 */
export type TranslationLanguage = string;

translationLanguages?: Record<RootLocale, TranslationLanguage>;
getTranslationLanguage(config, locale: RootLocale): TranslationLanguage;
getLocalesForTranslationLanguage(config, lang: TranslationLanguage | RootLocale): RootLocale[];
```

wdyt?

stevenle: ok done now.

jeremydw:
LGTM

PR #1230, opened 2026-06-12
chore: polish action logs

jeremydw:
- add "Show changes" button that opens the versions
- show slightly humanized tooltips (for actions that happen on the same day)
- slightly fix the alignment of the avatar cells

### DIFF
--- a/packages/root-cms/ui/components/ActionLogs/ActionLogs.tsx
+++ b/packages/root-cms/ui/components/ActionLogs/ActionLogs.tsx
@@ -652,8 +652,8 @@ function toTimestamp(ts: any): Timestamp | null {
  * tooltip showing the exact date and time.
  *
  * Formats:
- * - `compact` (default): `jun12 07:51am`
- * - `long`: `June 12, 2026 07:51am`
+ * - `compact` (default): `Jun 12, 10:21 AM`
+ * - `long`: `June 12, 2026 10:21 AM`
  */
 function ActionTimestamp(props: {
   timestamp: Timestamp;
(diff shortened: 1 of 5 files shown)


stevenle:
https://rootjs-dev--pr-1230-w09jv5ql.web.app

the new time formatting doesn't look like an improvement to the ux imo, it requires an extra mental step from me to figure out what time that change was made if i needed to figure out whether that lined up with another event outside of the system.

current:
Screenshot 2026-06-12 at 5 53 26 AM

this pr:
Screenshot 2026-06-12 at 5 52 53 AM

not sure exactly what you're trying to solve, would it help to have it moved from military time to having am/pm?

jeremydw:
@stevenle I'm basically trying to solve the problem of "was this a very recent edit or was it something done a while ago"? For example, if multiple people are actively using the tool I want a stronger recency signal (e.g. "5m ago"). 

I guess I also (personally) just struggle with the conversion from MM/DD HH:YY to the current day. I audited a few tools (Issue Tracker, Google Docs, GitHub) and they seem to format the history dates like this:

```
June 12, 7:51 AM
```

(And only include the year if it doesn't match the current.) I guess this is slightly easier for me because I can more easily compare it to the system clock in the top right of the screen. :P

WDYT about that format?

stevenle:
the way the component is laid out, the date/time needs to have a fixed width format so that the cols are consistent. we also need the format to be as short as possible (minimal extra whitespace) so that there's more space for the log message.

can you do this:

instead of a 50/50 layout, match the collection width with what's used on the collection list page and the action log side would be variable width
update the format to something like `jun12 07:51am`
on the full action logs page (`/cms/logs`), you can include a longer time format (but it hsould still be consistent in width), so something like `June 12, 2026 07:51am`

for the "recent actions" on the home page, the year isn't super important to me since for an active project it would be clear enough when the change was made, but if you think it's problematic not including the year then maybe we detect if the most recent change is from another year and we would display the year only in that scenario.

jeremydw:
Done

image

image

stevenle:
with the increased width for the component i'm thinking we can use the slightly longer `MMM dd HH:MMpm` formatting instead.

for the meta id associated with the action, it looks a little weird to see a big gap between the action type and the meta id. i think my preference is to treat that as one column, but if you feel like it needs to be aligned, what do you think about right-adjusting the action type so that they're aligned on the right edge?

something like this?

Screenshot 2026-06-12 at 10 23 57

jeremydw:
@stevenle Your mock looks good to me, i can make the update. Do you mind if we use the typical space before AM/PM? So

```
Jun 11, 10:30 AM
Apr 21, 09:45 PM
```

(With comma? No comma?)

stevenle:
adding all 4 variations for comparison:

Screenshot 2026-06-12 at 10 23 57
Screenshot 2026-06-12 at 12 11 13
Screenshot 2026-06-12 at 12 11 39
Screenshot 2026-06-12 at 12 12 15

your proposal with comma seems fine.

stevenle (approved):

jeremydw:
image

Know this was merged, but here's what it looks like ^

PR #1234, opened 2026-06-12
feat: add --log flag to root build with progress indicator

jeremydw:
## Summary

Large SSG builds currently emit one log line per rendered page — a 10k-page site produces 10k+ lines, which bloats CI logs (and log streaming isn't free). This PR replaces the per-file lines with an adaptive progress indicator while keeping the useful information.

### Modes (`root build --log <mode>`)

- **`progress`** (default)
  - Interactive terminals: live single-line progress bar — page count, %, bytes written, elapsed, ETA.
  - Non-interactive (CI): one progress line per ~10% milestone, throttled to ≥5s apart (~10 lines total).
  - Both: final summary with page count, total output size, build time, and the **largest pages** (preserves the main diagnostic value of the old per-file output).
- **`verbose`**: previous behavior — one line per output file.
- **`quiet`** (or global `-q`): only the final summary; vite output reduced to warnings/errors.

### Example output (CI, progress mode)

```
html output:
  213/2126 pages (10%) · 71.2 MB · 5.1s · eta 45.8s
  428/2126 pages (20%) · 142.9 MB · 10.3s · eta 40.9s
  ...
  ✓ 2126 pages (716.4 MB) in 51.2s
  largest pages:
  549.86 kB  dist/html/index.html
  532.96 kB  dist/html/contact/index.html
  ...
```

### Implementation notes

- New `BuildProgress` reporter in `src/cli/build-progress.ts` (injectable clock/stream, unit tested — 10 tests).
- `--log` validated via commander `choices()`.
- On build errors, the in-place progress line is cleared before the error prints, and the summary is suppressed.
- `js/css output` section unchanged (small); hidden in quiet mode.
- The progress line is rendered with `\r` rewrite only when the stream is a TTY, so piped/CI output never contains control characters.

## Test plan

- `pnpm test` in `packages/root` (33 files, 179 tests) ✅
- e2e against `examples/starter`: default/verbose/quiet modes + invalid value handling ✅
- Verified non-TTY output contains no ANSI rewrite sequences ✅

### DIFF
--- a/.changeset/build-log-progress.md
+++ b/.changeset/build-log-progress.md
@@ -1,17 +1,5 @@
 ---
-'@blinkk/root': minor
+'@blinkk/root': patch
 ---
 
 feat: add `--log` flag to `root build` with progress indicator
-
-`root build` now shows a progress indicator by default instead of printing
-one line per output file. In interactive terminals this is a live progress
-bar (count, %, bytes written, elapsed, ETA); in non-interactive environments
-(e.g. CI) a progress line is printed at ~10% milestones, keeping logs to a
-handful of lines instead of one per page. A final summary lists the total
-page count, output size, build time, and the largest pages.
-
-- `--log progress` (default): progress indicator + summary.
-- `--log verbose`: one line per output file (previous behavior).
-- `--log quiet` (or the global `-q`): only the final summary; vite output is
-  reduced to warnings and errors.
--- a/.changeset/build-log-progress.md
+++ b/.changeset/build-log-progress.md
@@ -1,0 +1,17 @@
+---
+'@blinkk/root': minor
+---
+
+feat: add `--log` flag to `root build` with progress indicator
+
+`root build` now shows a progress indicator by default instead of printing
+one line per output file. In interactive terminals this is a live progress
+bar (count, %, bytes written, elapsed, ETA); in non-interactive environments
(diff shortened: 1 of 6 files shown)


jeremydw:
@stevenle This was the result of some extra profiling I did for the SSG builds. The AI suggested that the log streaming wasn't free (good suggestion) so I worked with it a bit to produce an alternative logging behavior for SSG builds.

The AI suggested changing the default to the new `progress` it introduced, LMK if you think we should change the default back to `verbose` or if you have any extra feedback.

stevenle:
code generally looks good. i assume you tested and the indicator looks good from a ux perspective.

while you're in the build cli files, can you add some log some sort of message for the number of threads used? that'll help us troubleshoot what `--threads=auto` is using if we run into add'l oom issues.

jeremydw:
Added a threads log line per the feedback — prints under "html output:" so it sits next to the progress output it explains. Examples:

```
threads: 8 workers x 4 pages/worker (--threads=8)
threads: 14 workers x 3 pages/worker (auto = min(cpu 31, pages 212, memory 14))
threads: in-process, 10 concurrent pages (--threads not set)
```

The auto form surfaces all three limits so the binding constraint is visible when troubleshooting OOM. Suppressed in `--log quiet`.

stevenle:
that feels a little verbose but we can tweak that later. lgtm.

PR #1241, opened 2026-06-15
feat: add pagination to asset library page

jeremydw:
- this extracts the pagination from action logs into a shared component
- reuse it on the asset browser page
- also add seed script to test lots of assets

image


stevenle:
if you have this branch open can you add a changeset?
```
feat: add pagination to asset library page
```

stevenle (approved):

PR #1253, opened 2026-06-17
feat: add compare viewports to cms preview

jeremydw:
## Summary

Enhances the document preview viewer to support comparing **multiple viewports at once**. The device selector is now a multi-select, and a new **Compare viewports** toggle renders one iframe per selected viewport side by side, preserving the existing scaling logic of the single preview.

## Changes

- **Multi-select viewports** — device state is now `devices: DeviceType[]`. The legacy single-device localStorage preference is migrated automatically.
- **"Compare viewports" (2-up) toggle** — a new checkbox next to the Expand button (`IconColumns2`). When enabled the device buttons add/remove viewports; when disabled it behaves as the original single-select.
- **Smart 2-up pairing** — enabling compare always shows at least two viewports so the toggle has a visible effect: mobile/tablet pairs with desktop, desktop pairs with mobile (full preview → mobile + desktop). Disabling restores whatever was selected beforehand.
- **Side-by-side rendering** — multiple iframes render in a flex row with a small gap and per-viewport resolution labels. A single uniform scale fits all selected viewports in the available area; for one viewport it reduces to the previous behavior. Iframes are keyed by slot index so switching viewports just resizes (no reload).
- **Highlighting** — node highlight messages are broadcast to all visible preview iframes.

## Testing

- Updated `DocumentPagePreviewBar` visual tests (new multi-select props + a new 2-up case) and regenerated the affected baselines.
- UI bundle builds cleanly; no type or lint errors.


stevenle:
the preview link lgtm. the separate button seem fine to me from a ux perspective.

stevenle:
can you remove the visual tests? fwiw i like the tests but png's tend to bloat git repos, i think i'd like a better long term solution for how to handle visual tests.

PR #1254, opened 2026-06-17
refactor: optimize root jsx renderer (minimal fast path + attribute lookup)

jeremydw:
## Summary

Continues the Root JSX SSR renderer perf work from #1252. Optimizes **both**
render modes of `renderJsxToString`, while producing byte-identical output to
the previous renderer:

- **`minimal` mode** — now consistently faster than `preact-render-to-string`.
- **`pretty` mode** — now meaningfully faster than the previous pretty renderer.

## Benchmark results

Measured on Node v24, fair interleaved harness (each renderer alternated per
round to cancel CPU thermal drift), across three representative page shapes.

### `minimal` mode — vs `preact-render-to-string`

| Shape | vs `preact-render-to-string` | vs previous Root renderer |
| --- | ---: | ---: |
| card-grid (~520 nodes, attribute-heavy) | **1.54x** | +26% |
| deep-nest (depth 40) | **1.75x** | +35% |
| text-heavy (60 paragraphs) | **1.34x** | +15% |

The in-repo `pnpm vitest bench test/jsx-render.bench.ts` (which runs preact first
in each block, so it is biased *toward* preact) independently confirms
1.54x / 1.70x / 1.11x.

### `pretty` mode — vs the previous pretty renderer

Pretty mode is compared against the **previous pretty renderer** (not preact),
since preact does not emit formatted HTML. Ratios use per-round paired
measurements to cancel thermal drift:

| Shape | vs previous pretty renderer |
| --- | ---: |
| card-grid (~520 nodes, attribute-heavy) | **1.14x** (+14%) |
| deep-nest (depth 40) | **1.11x** (+11%) |
| text-heavy (60 paragraphs) | **1.14x** (+14%) |

Note: pretty mode inherently does the newline/block-formatting work that minimal
mode skips, so it cannot match minimal's headroom — these changes remove
essentially all of the *removable* per-element overhead from the pretty path.

### Minimal-mode path (all byte-identical output)

1. **Dedicated minimal-mode fast path** (`mRender`/`mElement`/`mChildren`) that
   drops the pretty-mode bookkeeping (the `nlFlag` newline side-channel, the
   `inline` parameter, block-element detection, and the mixed-content scan).
   `renderComponent` is shared between modes via an `activeRender` /
   `activeChildren` indirection.
2. **Null-prototype `PROP_TO_ATTR`** — the common "no remapping needed" attribute
   lookup (`class`, `href`, `id`, `data-*`, `aria-*`) now misses without walking
   the `Object.prototype` chain. This is the largest single win for
   attribute-heavy trees.
3. **`renderAttrs` rewrite** — `null`/`undefined` checked first, frequency-ordered
   dispatch (string/number values first), and numeric attribute values skip
   escaping entirely (a stringified number never needs HTML-escaping).
4. **Inlined common child shapes** — string leaves, single host-element children,
   and single-child chains skip a recursive dispatch.

### Pretty-mode path (all byte-identical output)

5. **Combined per-tag classification map** — `renderElement` previously did up to
   four separate `Set.has` probes per element (void / always-block /
   raw-content / default-block). These static classifications are precomputed
   into one `Map<tag, bitmask>` lookup. Runtime-configured block elements
   (`options.blockElements`) are tracked in a small separate set that is
   `undefined` (and skipped) in the common case. This is the dominant pretty
   win.
6. **Removed dead `isPretty` guards** — `render` / `renderChildren` /
   `renderElement` / `hasMixedContent` are only reachable in pretty mode (minimal
   uses the `m*` fast path), so the always-true `isPretty` checks they carried
   were eliminated.
7. **Inlined hot paths** — `typeof`-first dispatch in `render`, inlined string
   leaves in the children loops and single-string-child in `renderElement`
   (preserving the exact `nlFlag` newline bookkeeping), and `isTextNode` inlined
   into `hasMixedContent`.
8. **Single-concat string assembly** in `renderElement`.

## Verification

- 67/67 unit tests in `jsx-render.test.tsx` pass.
- 6/6 integration tests in `html-transform.test.ts` pass.
- Output is byte-identical to the previous renderer across all tested shapes in
  both `minimal` and `pretty` modes, including the `options.blockElements`
  configuration path.
- `pnpm build` and Prettier are clean.

Also adds `test/jsx-render.bench.ts`, the head-to-head benchmark used to drive
this work (imports the renderer from source, so it reflects local edits with no
rebuild).

### DIFF
--- a/.changeset/optimize-jsx-renderer-minimal-path.md
+++ b/.changeset/optimize-jsx-renderer-minimal-path.md
@@ -2,4 +2,4 @@
 '@blinkk/root': patch
 ---
 
-refactor: optimize root jsx renderer (minimal fast path + attribute lookup)
+refactor: optimize root jsx renderer (minimal + pretty fast paths)
--- a/.changeset/optimize-jsx-renderer-minimal-path.md
+++ b/.changeset/optimize-jsx-renderer-minimal-path.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root': patch
+---
+
+refactor: optimize root jsx renderer (minimal fast path + attribute lookup)
--- a/packages/root/src/jsx/jsx-render.ts
+++ b/packages/root/src/jsx/jsx-render.ts
@@ -138,87 +138,97 @@ const BOOLEAN_ATTRS = new Set([
   'selected',
 ]);
 
-/** JSX prop name -> HTML attribute name. */
-const PROP_TO_ATTR: Record<string, string> = {
-  acceptCharset: 'accept-charset',
-  autoCapitalize: 'autocapitalize',
-  autoComplete: 'autocomplete',
-  autoFocus: 'autofocus',
-  autoPlay: 'autoplay',
-  charSet: 'charset',
-  className: 'class',
-  colSpan: 'colspan',
-  contentEditable: 'contenteditable',
-  crossOrigin: 'crossorigin',
-  dateTime: 'datetime',
-  disableRemotePlayback: 'disableremoteplayback',
-  encType: 'enctype',
-  formAction: 'formaction',
-  formEncType: 'formenctype',
-  formMethod: 'formmethod',
-  formNoValidate: 'formnovalidate',
-  formTarget: 'formtarget',
-  frameBorder: 'frameborder',
-  hrefLang: 'hreflang',
-  htmlFor: 'for',
-  httpEquiv: 'http-equiv',
-  inputMode: 'inputmode',
-  itemProp: 'itemprop',
-  itemRef: 'itemref',
-  itemScope: 'itemscope',
-  itemType: 'itemtype',
-  maxLength: 'maxlength',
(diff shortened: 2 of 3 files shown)


jeremydw:
@stevenle Changes seem reasonable although it creates a new code path targeted towards minimal mode. Seems like if it's truly this much faster we could use minimal mode for our SSR sites.

stevenle:
@jeremydw if the speed tradeoff is insignificant i think i still prefer pretty mode for SSR since it makes it easier to debug.

jeremydw:
Let me see if I can squeeze improvement out of the pretty mode too.

stevenle:
do you wanna do that in a different pr or this one? this one lgtm to merge.

stevenle:
lgtm, lmk if you're still trying to squeeze in more or if this is good to go.

jeremydw:
@stevenle this should be good to go

PR #1259, opened 2026-06-18
feat: add source/reference support for select fields

jeremydw:
## Summary

Adds **`source`** support to `select` (and `multiselect`) fields, letting a field's options be sourced from a field in another CMS document instead of (or alongside) a static `options` list. The stored value is still a plain string (`select`) / string array (`multiselect`).

```ts
schema.select({
  id: 'flag',
  source: {
    doc: 'GlobalModules/flags', // <collection>/<slug>
    field: 'flags',             // the (array) field to read
    valueKey: 'name',           // key on each item -> stored value
    helpKey: 'description',      // key on each item -> dimmed help line
  },
  creatable: true,              // optional pick-or-create
});
```

### Field changes (`packages/root-cms`)
- `SelectField` gains `source?: FieldValueSource` and `creatable?`. `MultiSelectField` inherits both.
- New `DocumentSource` / `FieldValueSource` types (`core/schema.ts`). `valueKey` / `labelKey` / `helpKey` are keys read from each item (dot-paths supported); the value stays a plain string, so generated `.d.ts` types are unchanged.
- Resolver util `ui/utils/field-source.ts` (`resolveFieldSource`) fetches the source doc via the existing doc cache and normalizes the `_array` storage format.
- `SelectField.tsx` / `MultiSelectField.tsx` async-resolve options, show a loading spinner, support `creatable`, and render per-option help text via a custom `itemComponent`.
- An inline "open source doc" icon button (tooltip `Open <docId>`) sits to the right of the field and deep-links to the source field: `/cms/content/<doc>?deeplink=fields.<field>`.

### Docs test bed (`docs/`)
A runnable example of the feature: a `GlobalModules` collection (feature flags), a small conditions system with an `IsFeatureFlag` condition whose `flag` field is the document-sourced picker, an `If` page-module template to exercise it in the editor, and a preview route that renders the flags table.

## Test plan
- `pnpm build:core` + `pnpm build:ui:bundle` in `packages/root-cms` (green)
- `pnpm exec vitest run ui/utils/field-source.test.ts` (6 passing)
- `pnpm exec root build --ssr-only` in `docs` (green)
- eslint clean on all changed files
- Includes a changeset.

### DIFF
--- a/.changeset/select-field-value-source.md
+++ b/.changeset/select-field-value-source.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: add `source` (document-sourced options) and `creatable` to the select field (also available on multiselect) — bind a select's choices to a field in another CMS document via `source: {doc, field, valueKey?, labelKey?, helpKey?}`, with optional pick-or-create, per-option help text (`helpKey`), and an inline "open source doc" icon button
--- a/.changeset/select-field-value-source.md
+++ b/.changeset/select-field-value-source.md
@@ -2,4 +2,4 @@
 '@blinkk/root-cms': patch
 ---
 
-feat: add `source` (document-sourced options) and `creatable` to the select field (also available on multiselect) — bind a select's choices to a field in another CMS document via `source: {doc, field, valueKey?, labelKey?, helpKey?}`, with optional pick-or-create, per-option help text (`helpKey`), and an inline "open source doc" icon button
+feat: add source/reference support for select fields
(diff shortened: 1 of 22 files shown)


jeremydw:
@stevenle Took a stab at this feature; let me know if you have feedback about the schema changes, naming, or UI. As an FYI, I added support for help text within the `schema.select` UI too for this,.

stevenle:
this generally looks good, can you make sure that the reference values are cached somehow so that if many fields use the reference, there's only 1 doc fetch for the lifetime of the route? i think the memoization occurs currently on the component level which means each field that uses this would make its own separate fetch (i could be wrong, and also firebase may have its own caching mechanism so this optimization may be unnecessary)

stevenle (approved):

PR #1260, opened 2026-06-18
fix: ignore global cmd+k search hotkey in richtext fields

jeremydw:
## Problem

In the CMS, the global Cmd+K spotlight search fired even while the user was actively editing a Lexical rich text field. Lexical binds Cmd+K to "insert link", so pressing Cmd+K to create a link also opened the global search, clobbering the link flow.

## Cause

Mantine's `SpotlightProvider` `shortcut` prop is implemented via `@mantine/hooks` `useHotkeys`, whose `shouldFireEvent` only ignores `INPUT`/`TEXTAREA`/`SELECT` elements. Lexical rich text fields are `contenteditable` divs, which aren't covered, so the global shortcut still fired.

## Fix

- Disable Mantine's built-in shortcut (`shortcut={null}`).
- Register a custom Cmd/Ctrl+K `keydown` handler that opens the spotlight via `openSpotlight()` but bails out when focus is inside an editable target — input, textarea, select, or any `contenteditable` (rich text) field via `target.isContentEditable`.

Now Cmd+K inside a rich text field only triggers Lexical's link insertion, while still opening global search everywhere else.

### DIFF
--- a/packages/root-cms/ui/components/GlobalSearch/GlobalSearch.tsx
+++ b/packages/root-cms/ui/components/GlobalSearch/GlobalSearch.tsx
@@ -1,5 +1,9 @@
 import './GlobalSearch.css';
-import {SpotlightProvider, SpotlightAction} from '@mantine/spotlight';
+import {
+  SpotlightProvider,
+  SpotlightAction,
+  openSpotlight,
+} from '@mantine/spotlight';
 import {IconSearch} from '@tabler/icons-preact';
 import {ComponentChildren} from 'preact';
 import {useEffect, useMemo, useState} from 'preact/hooks';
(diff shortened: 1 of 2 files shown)


packages/root-cms/ui/components/GlobalSearch/GlobalSearch.tsx
```diff
@@ -385,7 +434,7 @@ function GlobalSearchInner(props: {
   return (
     <SpotlightProvider
       actions={augmented}
-      shortcut={['mod + K']}
+      shortcut={null}
```
stevenle, line 437: this change seems weird, maybe just remove this line altogether?

jeremydw: https://v4.mantine.dev/others/spotlight/#:~:text=To%20disabled%20keyboard%20shortcuts%20set%20shortcut%3D%7Bnull%7D%3A

```
To disabled keyboard shortcuts set shortcut={null}:
```

Maybe we need to leave it?

stevenle (approved):

PR #1261, opened 2026-06-18
feat: add compact document listing view

jeremydw:
## Summary

Adds a **Compact** view to the CMS collection document listing. The default listing renders large cards with a thumbnail, title, and preview URL — which is overkill for collections that don't have meaningful titles or metadata (e.g. a `Redirects` collection). The compact view condenses each row into a tabular layout.

## What's included

**Compact table** — a condensed, tabular listing with a header row and columns:
- Image (small thumbnail)
- ID
- Title
- Status (badges)
- Created / Modified (relative time, avatar + full date in tooltip)
- Three-dots actions menu

**View toggle** — a button between the **Sort** dropdown and the **New** button switches between the default and compact views. The choice is remembered **per-collection** in local storage.

**Schema override** — collections can force the compact listing:

```ts
export default schema.collection({
  name: 'Redirects',
  viewOptions: {compact: true},
  // ...
});
```

When set, the listing is locked to compact and the toggle is hidden.

**Sortable headers** — the ID, Title, Created, and Modified column headers are clickable to change the sort order, kept in sync with the sort dropdown (which gains matching **Title (A-Z/Z-A)** and **Least recently modified** options). Clicking toggles direction with a chevron indicator.

**Shared `UserActionTooltip`** — a new component renders the acting user's avatar alongside a timestamp and a "by &lt;email&gt;" line (capped width, wraps long emails, no avatar ring, and selectable text via `allowPointerEvents`). Used by both the compact Created/Modified cells and the doc status badges.

**Responsive + reload UX** — long release badges keep full size and the Status cell scrolls; the whole table scrolls horizontally below a min-width on narrow viewports. Reloads (e.g. changing the sort) now dim the existing rows with a sticky spinner overlay instead of blanking the page.

## Implementation notes

- `core/schema.ts`: added `viewOptions.compact` to the `Collection` type.
- `core/app.tsx`: serialize `viewOptions` to `window.__ROOT_CTX.collections`.
- `ui/hooks/useDocsList.ts`: added `title`/`titleDesc` (orders by the resolved `preview.title` field path, falling back to doc id) and `modifiedAtAsc` sorts.
- `ui/pages/CollectionPage/`: per-collection local-storage state, the toggle, sortable header cells, compact `DocsList` rendering, and reload overlay.
- `ui/components/UserActionTooltip/`: new shared tooltip component.
- `ui/components/DocStatusBadges/DocStatusBadges.tsx`: status badges now use `UserActionTooltip`.

## Testing

- `pnpm build:ui:bundle` and `pnpm build:core` succeed.
- `prettier` + `eslint` clean on changed files.

> Note: sorting by Title uses a Firestore `orderBy` on the title field, so docs missing that field won't appear while that sort is active — same behavior as the schema's existing custom `sortOptions`.

## Screenshot

image


stevenle:
dumb question: what's the `viewOptions: {compact: true},` config for? if it's a ui option it doesn't seem like one would need to opt-in to this new view type.

jeremydw:
> dumb question: what's the `viewOptions: {compact: true},` config for? if it's a ui option it doesn't seem like one would need to opt-in to this new view type.

It's to limit the UI to the compact view only. I was thinking it might be nice for the developer to have that fine-grained control so that they can force the small / compact view for collections that are, like, config-only (e.g. collections of redirects or config-only files that don't have URLs, previews, etc.)

Maybe we use a different signal for this?

stevenle:
so if the compact view is in the schema definition, would the button be hidden? 

i'm just trying to figure out how to avoid user confusion, i think they may toggle between one collection to another and be confused about why one looks different than the other.

i'm also pondering whether this should be the default / only view option. i think in the older layout we needed the bigger cards to help give balance to the design but with the "surfaces" this tabular format seems fine. we would be able to remove the "sort" dropdown at the top (and enable sorting via clicking on table headers), and users can optionally define custom meta fields to display in the table.

let me think this one through for next week.

jeremydw:
To answer your Q: the button is visible in all scenarios, _unless_ `compact: true` is present, which forces the compact view. (The original idea is to force the compact view for collections where the large view provides no value...)

I personally prefer the compact view (most CMS use compact view too) but I feel like there's something nice about the big images, titles, etc. especially for users that are used to it -- and for some collections that don't have too many docs in them.

But I'm open to anything - as long as there's a way to eventually have a compact view (and also clean up the UI for collections that just have configs/long lists/etc.)

stevenle:
ok let's do this:

- next to the icon let's add a "View: " label (i'm getting old and never really like long list of icons where i have to hover them to understand what they do). when you click on the icon it should present a menu for the different density options, i'm thinking "comfortable" (https://tabler.io/icons?icon=baseline-density-medium) and "compact" (https://tabler.io/icons?icon=baseline-density-small) but open to suggetions here
- instead of "id" maybe let's just show the "slug" in the table view (this reduces the chance of text ellipsis)
- when you hover over the created/modified date can you make sure it tells you who made the change? (we can potentially put avatars there but i think that might make the row too busy)
- i'm thinking the density setting should be "sticky" as the user navigates from one collection to another, and any collections with the schema setting would "lock" the view to that view only (menu greyed out or hidden altogether)

stevenle:
this mostly looks good, column widths are a little off esp when docs are in a release:

Screenshot 2026-06-23 at 10 24 53

jeremydw:
@stevenle I was thinking about this this morning - maybe we create a separate column for Release (but only show the "Release" column if any of the docs in the listing have a release)?

stevenle:
yeah that's fine for the compact view. we may have other badges there in the future but we can account for those when it comes up.

stevenle (approved):

PR #1262, opened 2026-06-18
feat: add renderMode option to override JSX render mode per request

stevenle:
## Summary
This PR adds support for overriding the JSX render mode on a per-request basis, allowing individual routes to use a different render mode than the default configured in `root.config.ts`.

## Key Changes
- **Added `renderMode` option to `HandlerRenderOptions`**: Routes can now pass `{renderMode: 'minimal' | 'pretty'}` to `ctx.render()` to override the default mode for that specific request
- **Updated `Renderer` class**: Modified `render()`, `renderHtml()`, and `renderJsx()` methods to accept and propagate the `renderMode` option through the rendering pipeline
- **Enhanced `getJsxRenderFn()`**: Now accepts an optional `renderOptions` parameter to apply per-render mode overrides while maintaining backward compatibility with the config-level default
- **Added comprehensive tests**: New test file validates that routes can override the render mode, with fixtures demonstrating both default and overridden behavior

## Implementation Details
- The `renderMode` parameter flows through the rendering pipeline: `render()` → `renderHtml()` → `renderJsx()` → `getJsxRenderFn()`
- Per-render mode overrides take precedence over the `jsxRenderer.mode` configured in `root.config.ts`
- The implementation maintains backward compatibility—if no `renderMode` is provided, the configured default is used
- Test fixtures include a config with `mode: 'pretty'` and routes demonstrating both default behavior and mode override via the handler

https://claude.ai/code/session_01YKwfcoGqqJKxbYNZaXhJA1

### DIFF
--- a/.changeset/red-words-knock.md
+++ b/.changeset/red-words-knock.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root': patch
+---
+
+feat: add renderMode option to override JSX render mode per request
--- a/packages/root/src/core/types.ts
+++ b/packages/root/src/core/types.ts
@@ -6,7 +6,7 @@ import {
 } from 'express';
 import {ComponentType} from 'preact';
 import {ViteDevServer} from 'vite';
-import {JsxRenderOptions} from '../jsx/jsx-render.js';
+import {JsxRenderMode} from '../jsx/jsx-render.js';
 import {Hooks} from '../middleware/hooks.js';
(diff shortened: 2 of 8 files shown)


jeremydw:
LGTM - I was going to say we could just abandon this given the results of our profiling exercise yesterday, but it might be nice to allow custom callsites to override the render mode when using the `Renderer` directly (unless you disagree - in which case we can abandon).

PR #1265, opened 2026-06-23
feat: add multiplayer profile images next to fields

jeremydw:
This PR slightly improves the "multiplayer" UI by adding profile images adjacent to the field each person is using.

- It also slightly tweaks the "focused" field UI, to reduce the horizontal padding and linkify the field name instead of the invisible anchor.

## Screenshots

Rendered from the real components (Preact + the actual CSS) in the Playwright visual-test harness, with Firestore presence data mocked — not live CMS captures.

**Presence avatars next to fields.** Other viewers focused on a field appear in the field header, and the field's input is outlined in that user's color. The outline is drawn outside the control's own border, so presence appearing/disappearing never reflows the field.

field presence before and after

**Topbar avatars.** The color ring is drawn just inside the photo's edge, so color-coding leaves the avatar's width/height unchanged.

topbar avatars before and after

**Inactive (disconnected) viewers** fade with `brightness()`/`saturate()` rather than `grayscale()`, which erased the user's color, and active avatars stack above their neighbors. Middle panel is the previous revision of this branch, for comparison.

inactive viewer treatment

**Focused field**, with the reduced horizontal padding.

focused field before and after

### DIFF
--- a/packages/root-cms/ui/components/DocEditor/DocEditor.css
+++ b/packages/root-cms/ui/components/DocEditor/DocEditor.css
@@ -27,6 +27,7 @@
 }
 
 .DocEditor__statusBar__saveState,
+.DocEditor__statusBar__viewers,
 .DocEditor__statusBar__statusBadges {
   margin-right: 16px;
 }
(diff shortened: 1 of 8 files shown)


stevenle:
thanks, this is pretty cool. a couple of notes:

- the inactive state is a little off when using opacity, maybe switch to brightness or something else (or re-order z-index so the active ones are on top):

Screenshot 2026-06-24 at 4 23 20 PM

- in figma, you can click on an avatar to follow a user, i wonder if we should add a click to deeplink to the field the user is looking at (if they are looking at a field)

- i wonder if we should also color code the users and highlight the field borders with the color matching the user (this might be a little much / too busy, so consider this feedback optional)

stevenle:
@jeremydw is this one ready for me to review?

jeremydw:
@stevenle - Yeah, sorry for the delay. You should be good to re-test this now

stevenle:
@jeremydw we may have some merge conflicts with https://github.com/blinkk/rootjs/pull/1311 - do you want me to submit mine first so we can rebase yours on top?

jeremydw:
yeah go ahead and do that

stevenle:
k my pr is merged, let's get yours rebased and then i can do a quick review.

stevenle:
some feedback:

- i think this PR increases the size of the avatars on the topbar? it feels a lot bigger than before. can you preserve the original width/height?
- the greyscale() style doesn't seem right, i thing brightness() may be more correct?
- per previous feedback, i think clicking on the avatar should deeplink to the field that the user is on

PR #1268, opened 2026-06-24
chore: tune the ui of the status badges

jeremydw:
the basic problem this fixes is space saving in the document list

- consolidate "draft + published" into a single "changed" badge to save horizontal space
- tune the ui to use smaller badges to save space
- add more detailed tooltips and clean up the tooltips
- add proper truncation when the release badges are too wide for their container
- clicking the status badges now shows the version history or diff (depending on the badge)

this mainly aims to clean up a ui like this:

image

with a proposed ui like this:

image

image

### DIFF
--- a/packages/root-cms/ui/components/CompareDraftModal/CompareDraftModal.css
+++ b/packages/root-cms/ui/components/CompareDraftModal/CompareDraftModal.css
@@ -1,0 +1,3 @@
+.CompareDraftModal__diff {
+  margin-top: 4px;
+}
--- a/packages/root-cms/ui/components/CompareDraftModal/CompareDraftModal.tsx
+++ b/packages/root-cms/ui/components/CompareDraftModal/CompareDraftModal.tsx
@@ -1,0 +1,62 @@
+import './CompareDraftModal.css';
+
+import {ContextModalProps, useModals} from '@mantine/modals';
+import {useModalTheme} from '../../hooks/useModalTheme.js';
+import {DocDiffViewer} from '../DocDiffViewer/DocDiffViewer.js';
+
+const MODAL_ID = 'CompareDraftModal';
+
+export interface CompareDraftModalProps {
+  [key: string]: unknown;
+  docId: string;
+}
+
+export function useCompareDraftModal(props: CompareDraftModalProps) {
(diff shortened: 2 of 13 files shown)


jeremydw:
@stevenle Could you take a look and send any feedback? The biggest non-visual change is the creation of a new "Changed" status (to save room primarily, and secondarily to be consistent with other systems).

By the way, I chose a shield icon for "Locked" because some people get confused and think "Locked" means no edits are allowed (it's more like publish protection or an embargo, though).

stevenle:
this is too big of a ui deviation, let's revert this PR or get it looking closer to the original.

Screenshot 2026-06-24 at 12 22 32

Screenshot 2026-06-24 at 12 22 39

jeremydw:
@stevenle 

Ready for another look - https://rootjs-dev--pr-1268-yphhldr1.web.app/cms/content/Pages

I reverted the coloring/style changes and eliminated the "changed" badges; kept the tooltip polish, and new "Show unpublished changes" modal, and the fixes for the alignment between the badges and column sizing.

The compact view is still difficult when you have a lot of badges visible and a small screen size but maybe we address that in a separate PR

stevenle:
this is looking better, thanks.

i'm wondering whether we should reverse our decision about having releases be its own column. when the statuses column has a few badges, the releases column creates more room than is necessary, example:

Screenshot 2026-06-24 at 15 02 11

one thing we could potentially do is collapse the releases badge from the id to instead be "release" if there's only 1, although i'm not sure what the best way to handle is if there are two or more releases (maybe having it say "2 releases" and having it link to a filtered list view of the releases page).

stevenle:
in the case of a lot of badges, would you be opposed to having them wrap to the next line (same with the title column)? that would create more room for smaller viewports, although traditionally our css hasn't accounted much for non-desktop sizes.

stevenle:
@jeremydw is this ready for me to re-review?

jeremydw:
@stevenle Just confirmed this should be good for another look; here's a screenshot of what it looks like

image

stevenle:
screenshot lgtm! will do a quick review and then merge.

PR #1269, opened 2026-06-24
feat: allow routes to override site-wide i18n.locales

stevenle:
## Summary
This PR adds support for individual routes to override the site-wide `i18n.locales` configuration, enabling more granular control over which locales are generated for specific routes during the SSG build process.

## Key Changes
- **Added `RouteConfig` interface** (`src/core/types.ts`): New interface that allows routes to export a `config` object with a `locales` property to override the site-wide i18n configuration
- **Updated `RouteModule` type** (`src/core/types.ts`): Added optional `config?: RouteConfig` property to allow routes to export configuration
- **Modified router logic** (`src/render/router.ts`): Updated route registration to check for route-level `locales` configuration and use it instead of the site-wide locales when building localized paths. The default-locale path is always generated regardless of route-level overrides
- **Added comprehensive test** (`test/i18n-route-locales.test.ts`): New test suite verifying that routes can successfully override locales and that only the specified locales generate localized paths
- **Added test fixture** (`test/fixtures/i18n-route-locales/`): Complete fixture with root config (en, fr, de locales) and two routes:
  - `index.tsx`: Uses site-wide locales (default behavior)
  - `limited.tsx`: Overrides to only generate French locale paths

## Implementation Details
- Route-level locale overrides are read from `routeModule.config?.locales` during route registration
- Falls back to site-wide `locales` if no route-level override is specified
- The default-locale path (without `[locale]` placeholder) is always generated for all routes, even those with locale overrides
- Changes apply to both regular routes and pod-based routes

https://claude.ai/code/session_01TgrTTgu1XuVQbhFqpwwjun

### DIFF
--- a/packages/root/src/core/types.ts
+++ b/packages/root/src/core/types.ts
@@ -179,12 +179,27 @@ export type GetStaticContent = (
   props: any
 ) => Promise<StaticContentResult | string> | StaticContentResult | string;
 
+/**
+ * Route-level config that a route module can export to override the site-wide
+ * serving behavior on a per-route basis.
+ */
+export interface RouteConfig {
+  /**
+   * Overrides the site-wide `i18n.locales` for this route. When set, the SSG
+   * build only generates localized paths for the locales listed here instead of
+   * the locales defined in `root.config.ts`. The default-locale path is always
+   * generated. Defaults to the site-wide `i18n.locales`.
+   */
+  locales?: string[];
+}
+
 export interface RouteModule {
   default?: ComponentType<unknown>;
   getStaticPaths?: GetStaticPaths;
   getStaticProps?: GetStaticProps;
   handle?: Handler;
   getStaticContent?: GetStaticContent;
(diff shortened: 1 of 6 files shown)


jeremydw (approved):
beautiful surgical change - ty. 

notably this does nothing for the CMS UI when it comes to choosing translations or locales, but that's sort of overloaded right now (esp with the v2 refactor looming), so let's scope this as-is and we can address any UI fixes down the road separately. (i can translate the content on the backend regardless of the UI).

stevenle:
re: translations, there's an option to exclude locales from translations and there's also a new translations map configuration that allows you to specify aliases between a "root locale" and a "translation id". as for toggling the locale in the cms ui i think in that scenario you should use the main locales and localization groups configurations to organize the locales, i don't think we want collections to have separate collection identifiers because that makes it difficult for the ui to know exactly where to grab locale ids from.

PR #1271, opened 2026-06-24
ci: retry firebase deploy on transient auth failures

jeremydw:
## Problem

Both deploy workflows intermittently fail at the `firebase deploy` step with:

```
Error: Failed to authenticate, have you run firebase login?
```

This is **not** a code/build regression — it's a flaky auth failure. The same commit both succeeds and fails depending on when it runs:

- `bef3c53`: succeeded at 19:47 & 19:58, failed at 20:42
- `45480ad`: failed at 18:50, succeeded at 19:10

`google-github-actions/auth@v2` reports "Successfully authenticated" and writes the credentials file, but firebase-tools (`lib/requireAuth.js` → `autoAuth()` → google-auth-library `GoogleAuth.getAccessToken()`) performs a network token exchange that occasionally fails transiently (it throws ~2s in, well under its own 15s timeout), surfacing the generic "Failed to authenticate" message.

## Fix

Wrap the Firebase deploy commands in a 3-attempt retry loop (15s backoff) in both deploy workflows:

- `.github/workflows/docs.yml` — "Deploy rootjs.dev to Firebase"
- `.github/workflows/docs-preview.yml` — "Deploy preview function to Firebase" and "Deploy preview to Firebase Hosting" (retry preserves the `--json` capture)

Note: `docs-preview.yml` is triggered by `issue_comment`, so the retry only takes effect for `/preview` runs once merged to `main`.

### DIFF
--- a/.github/workflows/docs-preview.yml
+++ b/.github/workflows/docs-preview.yml
@@ -330,26 +330,10 @@ jobs:
         run: |
           # Deploy only the PR-specific SSR function generated above. This does
           # not deploy or overwrite the public www-server or any other function.
-          #
-          # firebase-tools resolves credentials from GOOGLE_APPLICATION_CREDENTIALS
-          # via GoogleAuth.getAccessToken(), which performs a network token
-          # exchange that intermittently fails with "Failed to authenticate, have
-          # you run firebase login?" (the same commit succeeds on a re-run).
-          # Retry the deploy a few times to ride out those transient failures.
-          attempt=1
-          max_attempts=3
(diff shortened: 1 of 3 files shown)


stevenle:
wondering if it might be worth moving this to a deploy.sh script or something similar to avoid having so much logic contained in the `.yml` file.

jeremydw:
@stevenle Updated and moved it to a deploy_firebase script

stevenle (approved):

PR #1285, opened 2026-07-01
fix: remount sidebar tool iframe when switching tools

stevenle:
## Summary

Follow-up to #1279 (sidebar tool ↔ CMS URL sync). Fixes a regression and adds test coverage for the URL logic.

### Regression: switching tools showed the wrong tool
Navigating between two tools (e.g. `/cms/tools/a` → `/cms/tools/b`) kept showing the first tool. `SidebarToolsPage` matches a single route (`/cms/tools/:id`), so preact-iso keeps it mounted across the route change and only updates the `id` prop. The iframe's initial `src` is captured once in a `useState` initializer, so it never updated for the second tool.

**Fix:** key `ToolIframe` by tool id so it remounts (with the new URL, and a fresh sync effect) when the tool changes.

### Trailing-slash handling (with tests)
Extracted the URL math into `ui/utils/iframe-url-sync.ts` and added unit tests. Home detection now compares fully-resolved URLs, so a trailing-slash difference between the configured `iframeUrl` and the loaded location (`/tool` vs `/tool/`) is treated as the same home location and doesn't leave a stray `?path=` param. Restore-on-refresh already worked either way since the stored path is origin-relative; there are now tests asserting both.

## Testing
- `vitest run ui/utils/iframe-url-sync.test.ts shared/embed-protocol.test.ts` — 19 passed (tool switching, trailing slash w/ and w/o, query/hash preservation, relative iframe URLs, protocol guard).
- eslint clean on changed files; UI esbuild bundle builds successfully.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YR2YcPG6f9prt6VaSLgJb4)_


stevenle:
@jeremydw lmk if this fixes the issue, it keys off the tool path so that switching between `/cms/tools/a` and `/cms/tools/b` properly remounts the component.

jeremydw:
@stevenle Not sure if it's fixed; I patched the changes into our monorepo and it's still not working; and it even shows in the URL bar now `http://localhost:4007/cms/tools/design?path=%2Ftools%2Finsights%2F%3Fpreview%3Dtrue` instead which doesn't seem right :|

Maybe we need to update the docs site with two example tools mounted at /cms/tools/a and /cms/tools/b and we can test it locally there.

stevenle:
just prompted for a fix, will add a test in docs/ to make it easier to verify.

stevenle:
@jeremydw this is ready for another look

stevenle:
@jeremydw reminder about this one, there should be a test in the docs/ project for you to verify

jeremydw:
@stevenle Just tested it - it works! let's merge

PR #1293, opened 2026-07-09
feat: add middleware for requiring cms login

stevenle:
Adds a `requireCmsLogin()` express.js middleware that enforces login without the use of `?preview=true`. This PR also ignores the use of `?preview=true` query param in the `/cms/tools/` routes.

### DIFF
--- a/.changeset/major-rooms-care.md
+++ b/.changeset/major-rooms-care.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: add middleware for requiring cms login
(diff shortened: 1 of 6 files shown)


jeremydw (approved):
LGTM - Do you think we can update/reword the comment to signal whether `?preview=true` gets _added_ to the route automatically or if the route will just serve (and require CMS login) without it?

stevenle:
the jsdoc already has this at the top of the middleware:

```
/**
 * Express middleware that enforces Root CMS login on a custom route, without
 * requiring the `?preview=true` query param.
```

if this isn't clear enough lmk what comment you'd prefer.

jeremydw:
yeah i saw that but does it redirect and _add_ the query param? (like some of our old plugins do. if not, maybe add "and without redirecting to add it."

stevenle:
hmm maybe to me the indication of `without requiring the ?preview=true query param` is clear enough that it won't add it. presumably someone using this middleware wouldn't care one way or another?

PR #1297, opened 2026-07-10
fix: prevent the CMS loading screen from hanging indefinitely

stevenle:
## Summary
This PR adds timeout protection and error handling to prevent the CMS loading screen from hanging indefinitely when Firebase Auth or Firestore backends are slow, stuck, or require reauthentication.

## Key Changes

**Backend (plugin.ts)**
- Added `GET_CURRENT_USER_TIMEOUT_MS` (15s) timeout for `getCurrentUser()` calls to prevent requests from hanging when Firestore gRPC channels get stuck
- Implemented `withTimeout()` utility to wrap promises with deadline enforcement
- Added `isGcloudReauthError()` and `logGcloudReauthHint()` helpers to detect and handle Google Cloud credential expiration
- Moved static asset serving before auth middleware so the CMS shell can boot even when auth backend is unavailable
- Enhanced auth middleware to catch timeout errors and respond with explicit 503 status instead of hanging requests
- Consolidated duplicate ACL lookup logic by refactoring `getUserRole()` and `userExistsInAcl()` to use new `getUserAcl()` method

**Frontend (ui.tsx)**
- Added `AUTH_STATE_TIMEOUT_MS` (15s) watchdog timer to detect stalled Firebase Auth initialization
- Enhanced error screen with "Reload page" button for better UX when auth fails
- Added error callback to `onAuthStateChanged()` to handle auth state errors explicitly
- Wrapped `updateSession()` in error handler to prevent unhandled rejections

**Route Loading (lazy-route.tsx)**
- Added `ROUTE_IMPORT_RELOAD_GRACE_MS` (10s) grace period after triggering automatic reload
- Changed reload behavior to fall back to error screen if reload doesn't complete within grace period, preventing indefinite loading state

**Styling (global.css)**
- Added `.bootstrap__error-button` styles for the new reload button on error screens

**Testing (lazy-route.test.tsx)**
- Added explicit cleanup between tests to prevent test isolation issues
- Added test case for reload grace period timeout scenario

## Implementation Details
- Timeouts are applied at multiple layers: server auth verification (15s), client auth state (15s), and route reload grace period (10s)
- Static assets are now served without auth to ensure the CMS UI shell always loads
- Errors are logged with helpful hints for Google Cloud reauthentication issues
- The solution gracefully degrades: if auth eventually succeeds after timeout, the app still renders and recovers

https://claude.ai/code/session_01FUuTL9vvYTVcjfJRGWjSw2

### DIFF
--- a/.changeset/proud-hoops-tap.md
+++ b/.changeset/proud-hoops-tap.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+fix: prevent the CMS loading screen from hanging indefinitely
--- a/packages/root-cms/core/client.ts
+++ b/packages/root-cms/core/client.ts
@@ -1749,11 +1749,16 @@ export class RootCMSClient {
   }
 
   /**
-   * Gets the user's role from the project's ACL.
+   * Looks up the user's entry in the project's ACL with a single Firestore
+   * read. `exists` is `true` if the user (or their domain's wildcard entry,
+   * e.g. `*@example.com`) is in the ACL; `role` may be `null` for entries
+   * without an assigned role.
    */
-  async getUserRole(email: string): Promise<UserRole | null> {
+  async getUserAcl(
+    email: string
+  ): Promise<{exists: boolean; role: UserRole | null}> {
     if (!email) {
-      return null;
+      return {exists: false, role: null};
     }
     const docRef = this.db.doc(`Projects/${this.projectId}`);
     const snapshot = await docRef.get();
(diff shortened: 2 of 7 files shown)


jeremydw:
@stevenle Just flagging my specific error was on the /cms page _after_ the UI has fully loaded, so hopefully this covers that too!

stevenle:
the PR has a timeout for the firestore loading indicator and a reload button

jeremydw:
SG - As long as it handles the attached route (within `Loading__main`) we should be good. (I think it does, after re-reading the bullets).

Here's a screenshot of the exact place I've been getting stuck lately:

image

jeremydw (approved):

jeremydw:
Sadly on 3.1.8 this screen is still hanging for me sporadically - 
image

Refreshing the page fixes it. @stevenle Want to prompt the agent in the same session w/ this screenshot and PR comments and see if it can make a fix?

PR #1306, opened 2026-07-16
fix: send CSV imports as plain text instead of multipart/form-data

jeremydw:
## Problem

The CMS translations CSV import (`LocalizationModal`) uploads the selected file to `/cms/api/csv.import` as `multipart/form-data`. Some production deployments sit behind WAFs/edge proxies that block `multipart/form-data` POSTs outright, so the request is rejected before it ever reaches the server and the import fails with an opaque error.

## Change

- `csv.import` now accepts a JSON body (`{"csv": "<csv text>"}`) in addition to the existing `multipart/form-data` upload (kept for backwards compatibility).
- The CMS UI reads the selected file client-side (`file.text()`) and sends it as JSON.
- `/cms/api/csv.import` is included in the routes that use the larger (4mb) `body-parser` JSON limit, matching the AI prepare routes, so larger translation CSVs are not rejected by the default 100kb limit.

## Notes

- CSV files are text, so `file.text()` + JSON encoding is lossless.
- No API-consumer breakage: existing multipart clients continue to work; the response shape is unchanged.

### DIFF
--- a/.changeset/tidy-poets-shake.md
+++ b/.changeset/tidy-poets-shake.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+Send CSV imports as JSON instead of multipart/form-data. Some deployments sit behind WAFs/proxies that block multipart requests; the csv.import endpoint now accepts a JSON body (`{"csv": "..."}`) and the CMS UI sends JSON. Multipart uploads remain supported for backwards compatibility.
--- a/.changeset/tidy-poets-shake.md
+++ b/.changeset/tidy-poets-shake.md
@@ -2,4 +2,4 @@
 '@blinkk/root-cms': patch
 ---
 
-fix: send CSV imports as JSON instead of multipart/form-data
+fix: send CSV imports as plain text instead of multipart/form-data
--- a/.changeset/tidy-poets-shake.md
+++ b/.changeset/tidy-poets-shake.md
@@ -2,4 +2,4 @@
 '@blinkk/root-cms': patch
 ---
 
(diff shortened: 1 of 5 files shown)


.changeset/tidy-poets-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+Send CSV imports as JSON instead of multipart/form-data. Some deployments sit behind WAFs/proxies that block multipart requests; the csv.import endpoint now accepts a JSON body (`{"csv": "..."}`) and the CMS UI sends JSON. Multipart uploads remain supported for backwards compatibility.
```
stevenle: fix

jeremydw:
should be ready for another look now

packages/root-cms/ui/components/LocalizationModal/LocalizationModal.tsx
```diff
@@ -711,12 +711,14 @@ LocalizationModal.Translations = (props: TranslationsProps) => {
...
-        formData.append('file', file);
         try {
+          // Send the CSV as JSON rather than multipart/form-data; some
+          // deployments sit behind WAFs that block multipart requests.
+          const csv = await file.text();
           const res = await fetch('/cms/api/csv.import', {
             method: 'POST',
-            body: formData,
+            headers: {'Content-Type': 'application/json'},
+            body: JSON.stringify({csv}),
```
stevenle: current limit for this would be 4mb, csv uploads would fail for anything larger. do you anticipate any problems with this? if so, let's fallback to the multipart request for files larger than the limit.

jeremydw:
@stevenle - I changed this to use `text/plain` (or `text/csv`) and performs a client-side check against a 10 MB cap. If the user uploads a CSV that's larger it shows an error message. Thoughts?

stevenle (approved):

PR #1308, opened 2026-07-18
fix: prevent undefined values from reaching firestore

stevenle:
## Summary
This PR fixes a critical issue where `undefined` values could leak into Firestore `updateDoc()` payloads, causing write failures. The problem occurred when nested fields were modified after their parent object was queued for saving, or when remote snapshots arrived during pending removals.

## Key Changes

- **Added `removeUndefinedValuesDeep()` function** to recursively strip `undefined` values from update payloads before sending to Firestore. This handles the "paste race" scenario where a pasted array item's nested fields are modified before the parent is flushed.

- **Fixed snapshot merging** to properly apply pending `deleteField()` sentinels as actual deletions rather than copying the sentinel object into the data tree, preventing leakage to subscribers.

- **Updated `applyUpdates()` logic** to recognize `FieldValue` sentinels and delete keys instead of storing them as values.

- **Preserved store-object aliasing on retry** by keeping the original pending updates map around so failed flushes can re-queue updates without losing the shared references that later edits depend on.

- **Added bounds-checking in `arrayReducer`** to prevent stale indices from writing `undefined` entries into `_array` during concurrent updates (e.g., `moveUp`, `moveDown`, `removeAt`, `updateItem`).

- **Updated `JsonTrieStore.setValue()`** to delete keys when set to `undefined` rather than storing the undefined value, since objects are shared by reference with Firestore updates.

- **Added comprehensive test suite** (`useDraftDoc.controller.test.ts`) covering the paste race, snapshot-echo race, and undefined-pruning scenarios.

## Notable Implementation Details

- The fix maintains backward compatibility by only pruning `undefined` from plain objects and arrays, leaving non-plain objects (like Firestore `Timestamp` or `FieldValue` sentinels) untouched.
- The solution leverages the existing shared-reference pattern between the local store and pending updates, using it as a feature rather than fighting it.
- Bounds-checking in array operations prevents race conditions where concurrent updates could invalidate indices before a dispatch lands.

https://claude.ai/code/session_01ERLWyGLHtgxrSB18wfC9LD

### DIFF
--- a/.changeset/tricky-tires-allow.md
+++ b/.changeset/tricky-tires-allow.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+fix: prevent undefined values from reaching firestore
--- a/packages/root-cms/ui/components/DocEditor/DocEditor.tsx
+++ b/packages/root-cms/ui/components/DocEditor/DocEditor.tsx
@@ -1123,6 +1123,13 @@ function arrayReducer(state: ArrayFieldValue, action: ArrayAction) {
       const data = state ?? {};
       const order = [...(data._array || [])];
       const key = order[action.index];
+      if (!key) {
+        // The item no longer exists (e.g. removed by a concurrent update
+        // before the dispatch landed). Writing would target a bogus
+        // `undefined` key.
+        console.error('updateItem: no item at index', action.index);
+        return state;
+      }
       const newValue = action.newValue ?? {};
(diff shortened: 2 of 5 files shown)


stevenle:
@jeremydw fyi apparently the reason for intermittent "undefined" errors was due to the race condition where you add an array item, go into the new array item and remove a child array item. the "undefined" value didn't get cleaned up because the editor didn't look for those values in the children. this should fix it.

PR #1313, opened 2026-07-21
feat: add email notifications service for cms action logs

stevenle:
## Summary
Adds a built-in email notifications service that allows CMS administrators to receive email alerts when actions occur in the CMS (document publishes, schema changes, etc.). Emails are queued in Firestore and delivered by the Root.js email service using the App Engine Mail API.

## Key Changes

- **New `emailNotifications()` service**: A configurable notification service that sends emails triggered by CMS actions with support for:
  - Action filtering via wildcard patterns (e.g., `doc.publish`, `release.*`)
  - Custom recipient resolution (static list or dynamic function)
  - Customizable email templates with `{placeholder}` token support and dot notation for nested values
  - Data transformation before template rendering
  - Optional expiration dates for queued emails
  - Integration with the Root.js email service for immediate delivery

- **New `RootCMSClient.sendEmail()` method**: Queues emails in the `Projects/${projectId}/Emails` Firestore collection with support for:
  - Plain-text and HTML bodies
  - Automatic plain-text fallback generation from HTML
  - Custom sender addresses
  - Email expiration handling
  - Optional notification of the email service for immediate processing

- **Template rendering utilities**:
  - `renderEmailTemplate()`: Renders `{placeholder}` templates with support for nested value access, type-aware stringification (dates as ISO strings, objects as formatted JSON), and optional HTML escaping
  - `htmlToPlainText()`: Converts HTML to plain text for email body fallback

- **Comprehensive test coverage**: Unit tests for template rendering, service configuration, action filtering, recipient resolution, and email queuing

- **Public API exports**: New types and functions exported from the plugin for use in CMS configurations

## Notable Implementation Details

- Wildcard pattern matching is case-insensitive and supports `*` (any characters) and `?` (single character) with regex compilation caching for performance
- HTML escaping is applied only when rendering HTML email bodies to prevent injection
- The email service notification is fire-and-forget with a 60-second timeout to avoid blocking email queuing
- Default email templates provide sensible defaults with action, user, timestamp, and metadata information
- Function-based templates bypass placeholder rendering for full control over email content

https://claude.ai/code/session_01N9u8JvmMrq5sCQscg6SH4j

### DIFF
--- a/.changeset/email-notifications-service.md
+++ b/.changeset/email-notifications-service.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: add email notifications service for cms action logs
--- a/packages/root-cms/core/client.sendEmail.test.ts
+++ b/packages/root-cms/core/client.sendEmail.test.ts
@@ -1,0 +1,193 @@
+import {RootConfig} from '@blinkk/root';
+import {beforeEach, describe, expect, it, vi} from 'vitest';
+
+// Mock Firebase Admin.
+vi.mock('firebase-admin/app', () => ({
+  getApp: vi.fn(),
+  initializeApp: vi.fn(),
+  applicationDefault: vi.fn(),
+}));
+
+vi.mock('firebase-admin/firestore', () => ({
+  getFirestore: vi.fn(),
+  Timestamp: {
+    now: vi.fn(() => ({toMillis: () => 1234567890})),
+    fromDate: vi.fn((date: Date) => ({
+      toMillis: () => date.getTime(),
+      toDate: () => date,
+    })),
+  },
+  FieldValue: {},
+}));
+
+// Mock project module.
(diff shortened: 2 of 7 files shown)


stevenle:
@jeremydw do you generally want to review these new features or would you prefer to provide feedback ex post facto via gh issues?

jeremydw:
@stevenle For features like this let's submit + do post-review feedback. I think I'm only "more opinionated" on product/UI updates that can help the stakeholder/user relationship rather than infra/plumbing work (fully trust you and the AI to figure out the best approach for those). Let's submit if you think it looks good.

stevenle:
cool yes, sg. with the apis it typically takes me a few cycles to get the interface feeling "right" so agree about post-review feedback. (part of the reason why i prefer to have this as opt-in at first is so we can freely change the interface and easily grep if we need to make backward incompatible changes)

PR #1326, opened 2026-07-27
feat: sync url changes across the cms preview panes when comparing viewports

### LINKED ISSUE #1322: When multi pane is enabled, a url path change in one pane should update the other panes
**Description**

When a user interacts with a preview pane in the multipane mode, any url changes should apply to all of the open viewport panes.

stevenle:
## Summary
When comparing multiple viewport sizes side-by-side in the document preview, URL changes (navigation, hash changes, SPA routing) in one pane are now automatically mirrored to all other panes. This keeps the comparison apples-to-apples by ensuring all viewports display the same page.

Fixes #1322

## Key Changes

- **New utility module** (`preview-url-sync.ts`): Extracted URL comparison and sync logic into testable helpers that:
  - Extract comparable URL keys from iframe locations (stripping the CMS's internal `preview` param)
  - Detect which pane initiated a URL change
  - Determine which other panes need to be navigated to match
  - Reconstruct full preview URLs with the `preview=true` param

- **Enhanced preview iframe tracking**: 
  - Added `iframeUrlKeys` ref to track each pane's current and last-known URL state
  - Implemented `getIframeUrlKey()` to safely read iframe locations (handles cross-origin and loading states)
  - Implemented `syncPreviewUrls()` to detect and mirror URL changes across panes

- **Polling for client-side navigations**: 
  - Added `URL_SYNC_POLL_MS` (400ms) interval to detect URL changes that don't fire load events (hash changes, SPA routing)
  - Load event listener still triggers sync for server-side navigations
  - Polling is cleaned up when preview panes unmount

- **URL bar synchronization**: 
  - Updated `updateIframeUrlBar()` to strip query params and hash from displayed URLs (prevents editors from accidentally copying preview-only params)
  - Mirrors the page shown in previews to the URL bar

- **Comprehensive test coverage**: Added 159 lines of unit tests covering URL key extraction, sync detection, and edge cases (unreadable panes, mid-load states, single pane, etc.)

## Implementation Details

The sync logic uses a "source pane" approach: when a pane's current URL differs from its last-known URL, that pane is treated as the source of the change, and all other panes are navigated to match (except those already showing that URL or already sent there). This handles both user-initiated navigation and CMS-initiated navigation without conflicts.

The solution gracefully handles edge cases like cross-origin navigation (location unreadable), panes mid-load (location temporarily null), and multiple viewports loading the same URL simultaneously.

https://claude.ai/code/session_01C1yHBeLQbVM7qAExuigtKy

### DIFF
--- a/.changeset/wicked-carrots-share.md
+++ b/.changeset/wicked-carrots-share.md
@@ -2,4 +2,4 @@
 '@blinkk/root-cms': patch
 ---
 
-feat: sync url changes across the cms preview panes when comparing viewports
+feat: sync page navigations across the cms preview panes when comparing viewports
--- a/.changeset/wicked-carrots-share.md
+++ b/.changeset/wicked-carrots-share.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: sync url changes across the cms preview panes when comparing viewports
--- a/packages/root-cms/ui/pages/DocumentPage/DocumentPage.tsx
+++ b/packages/root-cms/ui/pages/DocumentPage/DocumentPage.tsx
@@ -35,7 +35,6 @@ import {testCanEdit} from '../../utils/permissions.js';
 import {
   getPreviewPathFromUrlKey,
   getPreviewSrcFromUrlKey,
-  getPreviewSyncUpdate,
   getPreviewUrlKey,
   getPreviewUrlKeyFromUrl,
 } from '../../utils/preview-url-sync.js';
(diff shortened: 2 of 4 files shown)


stevenle:
@jeremydw this is a PR to keep multiple panes in sync so that if, for example, a feature flag is updated on one pane it updates the other panes.

question for you: i'm thinking about having it ignore hash changes and only respond to url path / query param changes. wdyt? do we need hashes synced for any reason, or is that possibly troublesome for certain types of ui frontend debugging?

jeremydw:
I can see us needing hashes synced but i think that adds extra complexity so I'd prob vote to do that later only if we identify it as a need

stevenle:
ok cool, i asked claude to respond only to server-side nav changes, which removes the polling + no longer responds to hash change.

PR #1328, opened 2026-07-28
fix: memory leak in the dev schema watcher that regenerates root-cms.d.ts

jeremydw:
## Summary

In dev, `cmsPlugin` regenerates `root-cms.d.ts` whenever a `.schema.ts` file changes by calling the CLI's `generateTypes()` in-process. Each call:

1. runs `loadRootConfig()`, which bundles `root.config.ts` to a temp module via `bundle-require` — a new entry in node's ESM cache **on every save, forever** (the ESM cache is never evicted), and
2. boots a standalone vite server via `viteSsrLoadModule()` to load the project's schema registry, then `close()`s it — but the server is **not fully released** after close. Heap snapshots of a long dev session showed dozens of complete `RunnableDevEnvironment`s / module graphs / plugin containers still alive, each pinned from a GC root (a native global handle through the rollup `onWarn` closure).

On a project with a few hundred schema files (where schema modules also transitively import `root.config.ts`), this leaked **~5MB of post-GC heap per schema save**, measured via CDP `HeapProfiler.collectGarbage` + `Runtime.getHeapUsage` while editing. A schema-heavy dev session eventually OOMs the dev server even with a raised `--max-old-space-size`.

The throwaway servers also share the project's `node_modules/.vite` dep-optimizer cache with the real dev server and stamp it with a different config hash, causing spurious `✨ optimized dependencies changed. reloading` full reloads (each followed by a slow cold SSR render) during normal browsing.

## Fix

Load the project module through the **running dev server's** module graph instead of booting a new vite server per change:

- `configureServer()` captures the dev `viteServer` (same `server.get('viteServer')` mechanism `configureServerPlugins` uses).
- The watcher invalidates the `virtual:root/schemas` registry module (so added/removed schema files are picked up — edits to existing files are invalidated by vite itself) and calls `viteServer.ssrLoadModule(project.js)`.
- `generate-types.ts` gains `generateTypesFromProject(project, rootDir)` so the dts generation is reusable; the CLI entry point (one-shot process, where the standalone server is fine) is unchanged.
- Regens are single-flight with trailing coalescing, so a burst of saves runs one regen at a time and always ends with a fresh pass.
- Failures (e.g. a schema saved mid-edit that doesn't parse) are caught and logged instead of surfacing as an unhandled rejection like before.

Since only the invalidated modules re-execute, regens through the live graph are near-instant.

## Verification

- Ran the `examples/cms` dev server on this branch: schema saves regenerate `root-cms.d.ts` through the running server (renamed field showed up in the output d.ts), bursts coalesce, and a syntax-error save logs and recovers on the next save.
- RSS flat across 20+ consecutive schema-save regens (previously each save leaked the config bundle + a pinned vite server).
- On the affected large project, post-GC heap grew 5MB per schema save before; with the watcher path bypassed (`watch: false`) heap is flat across the same edit loop — this PR makes the default watcher behave the same way while keeping types regeneration.
- `tsc --noEmit -p tsconfig.build.json` and `eslint` pass; `pnpm build:core` output verified.

## Notes

- `viteSsrLoadModule()` itself still leaks when called repeatedly in a long-lived process (the created server isn't fully released after `close()` — appears to be pinned by a native handle from the bundler binding). Worth a follow-up look since other in-process callers would hit the same thing; this PR removes the only per-file-change caller.


jeremydw:
cc: @andrespr

stevenle:
lgtm but can you confirm you tested this and it doesn't have any known regressions?

stevenle:
just tested locally, will merge now and then cut a release.

PR #1352, opened 2026-08-05
feat: add a `build.excludeElements` option for preview-only elements

### LINKED ISSUE #556: Exclude certain asset files from SSG builds
**Description**

Certain custom elements are intended for preview-only (e.g. help overlays, internal tools, etc.). For SSG builds, there should be some option for excluding certain elements (maybe excluding ones matching a certain prefix, e.g. `debug-foo`).

stevenle:
Fixes #556.

### Problem

Some custom elements are only meant for previews — help overlays, debug grids, internal tools — and shouldn't ship in a static build.

The existing `elements.exclude` can't express that. It matches file paths and applies to every command, so an excluded element disappears from the dev server too, which is exactly where it's supposed to work.

### Change

Adds `build.excludeElements`, matched against the element's **tag name** and applied only to the SSG phase of `root build`:

```ts
export default defineConfig({
  build: {
    excludeElements: ['help-overlay', /^debug-/],
  },
});
```

Matching elements are not bundled, their assets are not emitted to `dist/html`, and their `<script>` / `<link rel="stylesheet">` tags are not auto-injected into rendered pages.

The value is an `ElementTagNameMatcher` (exported from `@blinkk/root`):

```ts
export type ElementTagNameMatcher =
  | Array<string | RegExp>
  | ((tagName: string) => boolean);
```

Strings match exactly (`'debug-'` does not match `debug-panel`), RegEx patterns are tested against the tag name, and the two can be mixed in one list. The predicate form covers logic a pattern can't express — chiefly allowlisting a known set of elements, which otherwise needs a negative-lookahead regex.

**Where it does *not* apply:**

- **The dev server** — preview-only elements keep working while developing.
- **`root build --ssr-only`** — that flag generates the pre-built files for SSR mode, where pages are rendered on demand, so those elements are kept in the output.

### Implementation

The filter lives in `getElements()`, gated on a new `isSsgBuild` option, so every downstream consumer inherits it from one place. `build.ts` derives the flag from the existing `--ssr-only` local:

```ts
const elementGraph = await getElements(rootConfig, pods, {isSsgBuild: !ssrOnly});
```

Because the elements never enter the graph, they're dropped from the vite client input (never bundled), from `dist/html` (no asset copied), and from `dist/.root/elements.json` — so the build worker threads and the prod SSR server stay consistent with the SSG output. `dev.ts` passes no options and is unchanged.

Note this excludes the element's **assets, not its HTML**. A `<debug-panel>` tag rendered by a route or template is still present in the build output, it just never upgrades into a custom element.

### Notes

- `test/element-graph.test.ts` (new) covers RegEx, exact-string, mixed, and predicate matching, the partial-string negative case, and that `elements.exclude` still applies to every command.
- `test/elements.test.ts` adds an SSG case with an inline HTML snapshot showing no injected script, plus an `--ssr-only` case asserting the `debug-panel` asset *is* emitted and present in `elements.json`. The latter fails against the first commit on this branch, so it's a real regression test for the `--ssr-only` fix.
- Full `@blinkk/root` suite passes: 39 files, 223 tests. `eslint` and `tsc --noEmit` are clean on every file touched here; the pre-existing lint/tsc errors elsewhere in the package are untouched.
- Scoped to elements, matching the issue. `bundles/` is left alone since bundles are referenced explicitly via `<Script src="/bundles/main.ts">` — excluding one would leave a dangling asset reference rather than degrading cleanly. Happy to cover them too if you want it here.

---

Generated with Claude Code (Claude Opus 5).

### DIFF
--- a/.changeset/quiet-poets-repeat.md
+++ b/.changeset/quiet-poets-repeat.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root': patch
+---
+
+feat: add a `build.excludeElements` option for preview-only elements
--- a/packages/root/src/cli/build.ts
+++ b/packages/root/src/cli/build.ts
@@ -105,7 +105,7 @@ export async function build(rootProjectDir?: string, options?: BuildOptions) {
     }
   }
 
-  const elementGraph = await getElements(rootConfig, pods);
+  const elementGraph = await getElements(rootConfig, pods, {isBuild: true});
   const elements = Object.values(elementGraph.sourceFiles).map((sourceFile) => {
     return sourceFile.filePath;
   });
--- a/packages/root/src/cli/build.ts
+++ b/packages/root/src/cli/build.ts
@@ -105,7 +105,11 @@ export async function build(rootProjectDir?: string, options?: BuildOptions) {
     }
   }
 
-  const elementGraph = await getElements(rootConfig, pods, {isBuild: true});
+  // `build.excludeElements` drops preview-only elements from the SSG output.
+  // An `--ssr-only` build renders pages on demand, so it keeps them.
+  const elementGraph = await getElements(rootConfig, pods, {
+    isSsgBuild: !ssrOnly,
+  });
   const elements = Object.values(elementGraph.sourceFiles).map((sourceFile) => {
     return sourceFile.filePath;
   });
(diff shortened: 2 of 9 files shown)


stevenle:
@jeremydw FYI for our use case i think we would use the `(tagName: string) => boolean` signature for this config option and provide an utility function within the design system that returns true if the element is an "internal-only element".

jeremydw:
looks good. function that returns a bool is perfect

jeremydw (approved):

PR #1359, opened 2026-08-06
feat: add AI tools for reading Google Docs and Sheets

### LINKED ISSUE #1358: Add more Root AI tools (fetch Google Doc)
We've already got the ability to read Google Drive files via the "Paste from Google Drive" feature for file uploads and the export/import to Google Sheets translation feature.

I figure we should give Root AI the ability to fetch Google Docs/Sheets/Drive files as a tool. Then folks can say "Update the page with the content from this doc" (it should work for many situations)

Image

stevenle:
Adds three new read-only AI tools that let users point Root AI at Google Workspace content without copy/pasting:

- `gdoc_get` — reads a Google Doc as markdown
- `gsheet_get` — reads cell values from a Google Sheet tab
- `gdrive_getFile` — reads arbitrary Drive files as text when possible

Fixes #1358 

## Implementation

**Core tool definitions** (`core/ai-tools-google.ts`):
- Defines the three tools with Zod schemas and descriptions for the model
- Implements a pluggable `GoogleToolBackend` interface for the actual API calls
- Converts `GoogleToolError` exceptions into structured tool results so the model can explain failures to users
- Tools are read-only by design; no create/edit/share operations are exposed

**Browser backend** (`ui/components/RootAIChat/clientGoogleTools.ts`):
- Implements `GoogleToolBackend` using the Drive and Sheets REST APIs
- Calls APIs directly with the signed-in user's own OAuth token (no server-side credential storage)
- Handles URL/file ID parsing, file type detection, and text export (Docs → markdown, Sheets → CSV-like grid)
- Implements smart error mapping (401 → token refresh, 403 → permission denied, 404 → not found, etc.)
- Includes `maybeAuthorizeGoogle()` to prompt for sign-in when a message references a Google link, before tool execution begins

**Google OAuth helper** (`ui/utils/google-auth.ts`):
- Centralized token and consent management shared with existing CMS Google features (Drive picker, Sheets export)
- Tokens live in memory only; consent record persists in localStorage to skip repeat consent dialogs
- Supports both interactive (user gesture required) and silent token requests

**Integration**:
- Updated `RootAIChat` and `AiEditModal` to include Google tools in the tool set (empty when `gapi` is not configured)
- Updated system prompt generation (`core/ai.ts`) to describe the tools and their error handling
- Updated tool label formatting for UI display

The tools only work when the project configures `gapi: {apiKey, clientId}` on the cmsPlugin; otherwise the tool set is empty so the model never sees unavailable tools.

https://claude.ai/code/session_0184pa91z4pvtMu5AEYxzm4o

### DIFF
--- a/.changeset/tidy-moons-gather.md
+++ b/.changeset/tidy-moons-gather.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: add AI tools for reading Google Docs and Sheets
--- a/packages/root-cms/core/ai-tools-google.ts
+++ b/packages/root-cms/core/ai-tools-google.ts
@@ -1,0 +1,323 @@
+/**
+ * Google Workspace read tools exposed to the chat model when the project
+ * configures `gapi: {apiKey, clientId}` on the cmsPlugin.
+ *
+ * These let the user point Root AI at content that lives outside the CMS
+ * ("update the page with the copy from this doc", "add a row per entry in
+ * this sheet") without copy/pasting it into the chat.
+ *
+ * Like the CMS read tools in `ai-tools.ts`, the tools carry an `execute` that
+ * delegates to a pluggable backend. The browser supplies one that calls the
+ * Drive/Sheets REST APIs with the signed-in user's own OAuth token (see
+ * `ui/components/RootAIChat/clientGoogleTools.ts`), so a user can only read
+ * files they personally have access to.
+ *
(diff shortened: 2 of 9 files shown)


jeremydw:
Awesome - great that it handles the auth too. Only nitpicky feedback is on the tool names, wonder if they should have more namespacing, e.g. `integrations_google_docs_get` or something (in case we want to build on it more later). In any case the actual tool names are pretty internal so we can change it later if needed

jeremydw (approved):

PR #1369, opened 2026-08-12
feat: warn or block deleting and unpublishing docs with doc references

willthomson:
Adds an opt-in "reference guard" to the doc delete and unpublish flows. When enabled, the confirm modal checks the dependency graph for other docs that reference the doc and either warns the editor or blocks the action until the references are removed. Without this, deleting or unpublishing a still-referenced doc silently leaves dangling references and dead links on the live site.

## Usage

```ts
cmsPlugin({
  dependencyGraph: {
    referenceGuard: 'warn', // or 'block'; `true` is shorthand for 'warn'
  },
});
```

Requires the dependency graph. Disabled by default — the delete/unpublish modals are unchanged when the option is unset.

## Changes

- **Config**: Added `referenceGuard?: boolean | 'warn' | 'block'` to `CMSDependencyGraphConfig`, with a `resolveReferenceGuardMode()` normalizer alongside `resolveDependencyGraphConfig()`. The mode reaches the UI via a new `dependencyGraphReferenceGuard` field on `__ROOT_CTX`, next to the existing `dependencyGraphEnabled` flag.

- **API**: `dependency_graph.query` accepts a `direction: 'dependencies' | 'dependents'` param (default `'dependencies'`) and echoes it in the response, wiring up the previously uncalled `DependencyGraph.getDependents()` reverse lookup instead of adding a new endpoint.

- **UI**: `onDeleteDoc`/`onUnpublishDoc` in `DocActionsMenu` pre-fetch direct dependents before opening the confirm modal via a new `cmsCheckReferenceGuard()` util. Delete checks the union of the draft and published graphs; unpublish checks published only (the draft copy survives). Dependents render with `DocPreviewCard` (capped at 10, "+N more"), following the `ReferenceDocs` pattern in `PublishDocModal`. In `'block'` mode the confirm button is disabled.

- **Tests**: Coverage for the config resolver (core) and for `cmsCheckReferenceGuard()`'s request shapes, draft/published union, and fail-open paths (ui).

## Implementation Details

The check fails open: if the guard is disabled, no docs reference the doc, or the graph query errors, the existing confirm flow runs unchanged — a graph outage should never prevent edits. Only direct dependents (`transitive: false`) are shown, since those are the docs that actually contain the breaking reference. The modal notes that only reference fields are checked — hardcoded links in text fields aren't detected, and draft-mode edges can lag until the next cron scan.

Generated with Claude Code (Claude Fable 5).

### DIFF
--- a/.changeset/shiny-doors-listen.md
+++ b/.changeset/shiny-doors-listen.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: warn or block deleting and unpublishing docs that other docs reference
--- a/packages/root-cms/core/api.ts
+++ b/packages/root-cms/core/api.ts
@@ -358,7 +358,8 @@ export function api(server: Server, options: ApiOptions) {
   >();
 
   /**
-   * Queries the dependency graph for the docs referenced by one or more docs.
+   * Queries the dependency graph for the docs referenced by one or more docs,
+   * or — with `direction: "dependents"` — the docs that reference them.
    * Requires the `dependencyGraph` cmsPlugin option to be enabled.
    *
    * Authentication: any signed-in CMS user.
@@ -368,6 +369,13 @@ export function api(server: Server, options: ApiOptions) {
    * ```
    * POST /cms/api/dependency_graph.query
    * {"docIds": ["Pages/index"], "mode": "draft", "transitive": true}
+   * ```
+   *
+   * Reverse lookup:
+   *
+   * ```
+   * POST /cms/api/dependency_graph.query
+   * {"docIds": ["Pages/index"], "direction": "dependents", "transitive": false}
    * ```
    */
   server.use(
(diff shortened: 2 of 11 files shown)


stevenle:
i'm tempted to say this feature would be better as a first-party managed "check" and we build the checks flow into the publishing system similar to how "presubmits" work within google. checks already have "warn" and "error" states where the latter would prevent publishing. we would need to overhaul our publishing ui to enable checks flows, approvals, etc. so this one may require a few rounds of thought / ux experimentation.

stevenle:
fyi i just prompted claude with the following:

```
i'd like to add the ability to run required checks as a publishing step for docs and releases, configurable on a collection-by-collection basis. "required" checks should halt publishing if failed. for "warnings" the publishing should continue with the warning messages being presented to the user at the end.

there should be an option for admins to skip checks altogether when publishing. admins should also be able to bypass "failed" checks.

if you need any clarification, please ask.
```

once this is in i think we can convert your code to a first-party check and users can opt into it as needed. note: i think the dependency graph isn't widely used yet so we may have to modify your check to not rely on the dependency graph.

stevenle:
closing this PR in favor of "required publishing checks" which was implemented in:
https://github.com/blinkk/rootjs/pull/1371

this feature should be available in `v3.4.8`

PR #1374, opened 2026-08-17
fix: let admins reset a user's stale Google sign-in link

jeremydw:
## Problem

A user whose Google account is deleted and recreated keeps the same email address but gets a new federated id (`sub`). Projects on the default "one account per email address" setting have Identity Platform attach a federated identity to the account that already exists for that address — and it rejects the attach when that account already carries a `google.com` provider with a *different* federated id.

The result is that `signInWithPopup()` on the CMS sign-in page fails with:

```
Sign in failed: Firebase: Error (auth/provider-already-linked).
```

Two problems with that today:

1. The message is meaningless to the person reading it, and the code is misleading — the CMS never links providers, so the error looks like a CMS bug.
2. Nothing in the CMS can clear it. Recovering the account means someone with Firebase admin credentials unlinking the stale provider by hand (Firebase console or the Identity Toolkit API) every time it happens.

## Change

**`POST /cms/api/users.reset_sign_in` (ADMIN-only)** — unlinks the `google.com` provider from the account with the given email so that the next sign-in attaches the identity the user is actually signing in with. The Firebase uid, the user's role, and their content are untouched: CMS access is keyed off the email address (`getUserAcl()`), not the uid. The endpoint is a no-op (`{success: true, reset: false}`) when no account or no Google provider exists, and the action is written to the project's action log.

Requiring ADMIN keeps the capability strictly below what an admin already has — an admin can grant and revoke CMS access outright from the same page.

**"Reset sign-in" in Settings** — an action next to each user in the Share box, behind a confirm modal that explains what it does. This is the piece that removes the "file a ticket for someone with Firebase credentials" step. It stays out of the way: it renders only for admins, only on row hover or keyboard focus, and only for users who actually have a Google link to unlink.

**`POST /cms/api/users.sign_in_status` (ADMIN-only)** — batched `getUsers()` lookup (100 identifiers per call) backing that last filter. It reports `{exists, hasGoogleLink}` per email, which hides the action for `*@domain` wildcards and for users who were granted access but have never signed in. Whether an *existing* link is stale can't be answered server-side — that only surfaces when the user tries to sign in — so the action remains available for accounts that do have one.

**Plain-language sign-in errors** — new `shared/sign-in-errors.ts` maps the errors a user can act on (`auth/provider-already-linked`, `auth/network-request-failed`, `auth/popup-blocked`, `auth/unauthorized-domain`, `auth/user-disabled`) to messages that say what to do, and falls back to the raw message otherwise. The stale-link message points at the Settings action, so the user knows what to ask for.

## Alternatives considered

- **Automatic self-service repair from the sign-in page.** Attractive, but a user who can't sign in has no credential to prove ownership of the address: on this failure the Firebase SDK throws without a `_tokenResponse`, so there's no OAuth credential to hand to the server. That leaves an unauthenticated endpoint that mutates auth records on request, which isn't a trade worth making for a rare failure. A variant that re-obtains a Google ID token through GIS and verifies it server-side would work, but it needs an OAuth client id configured per project, so it can't be the general path.
- **Flagging the specific user who failed.** The sign-in page could report the failure so the action only appears for that user, but the error carries no identity: on this code the SDK throws through `_fail()` with no `_tokenResponse`, so the email is only known when the browser happens to have a last-sign-in hint. Left out rather than shipping a best-effort flag behind a public write endpoint.
- **Deleting the Firebase Auth user instead of unlinking.** Same end state for the CMS, but it throws away the uid, which projects sharing the Firebase project may key data off.

## Testing

- `packages/root-cms/shared/sign-in-errors.test.ts` (new, 6 cases) — `pnpm vitest run shared/` passes (13 files, 156 tests).
- `pnpm build` in `packages/root-cms` passes; `tsc --noEmit` reports no errors in the touched files.
- Verified the underlying fix out-of-band against a real locked-out account: unlinking `google.com` from the record restored sign-in, with no change to the user's role or content.
- The Firestore-emulator suite wasn't run locally (no JDK on this machine).

### DIFF
--- a/.changeset/lucky-pears-fetch.md
+++ b/.changeset/lucky-pears-fetch.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+fix: let admins reset a user's stale Google sign-in link
--- a/packages/root-cms/core/api.ts
+++ b/packages/root-cms/core/api.ts
@@ -2,6 +2,7 @@ import {promises as fs} from 'node:fs';
 import path from 'node:path';
 import {Server, Request, Response} from '@blinkk/root';
 import {multipartMiddleware} from '@blinkk/root/middleware';
+import {getAuth} from 'firebase-admin/auth';
 import {toTranslationLanguages} from '../shared/translation-languages.js';
 import {
   buildChatSystemPrompt,
@@ -51,6 +52,9 @@ const SYNC_PROXY_ALLOWED_HOSTS: RegExp[] = [
 
 /** Max response size the asset sync relay will forward (100MB). */
 const SYNC_PROXY_MAX_BYTES = 100 * 1024 * 1024;
+
+/** The only sign-in provider the CMS uses. */
+const GOOGLE_PROVIDER_ID = 'google.com';
 
 type DocVersion = 'draft' | 'published';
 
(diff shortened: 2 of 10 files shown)


jeremydw:
FYI; We had a situation where a user's Google Workspace account was deleted and then recreated. After that they were unable to login. This PR proposes a way for admins to reset the account from the UI.

Note that this adds the "reset" link feature (on hover, for admins) in the share box. We also can't show it conditionally without logging an error when a user tries to sign in and fails (via an unauthenticated request).

Open to UI suggestions on how this could be hidden further.

stevenle (approved):

PR #1379, opened 2026-08-20
feat: warn and recursively delete non-empty asset folders

stevenle:
Deleting a folder in the asset library that still had contents failed with a `Folder is not empty.` error, leaving users to empty the folder by hand before it could be removed. The delete confirmation now surfaces this as a warning up front and lets the user proceed with a recursive delete.

## Changes

**`ui/utils/assets.ts`**

- `deleteAsset()` takes a new `options.recursive` flag. When set, every asset nested below the folder (at any depth) is deleted along with it, in batches of 400 writes. Without it, a non-empty folder still refuses to delete — now throwing a typed `FolderNotEmptyError` instead of a bare `Error`.
- New `countFolderContents()` returns the number of nested files and folders, so the UI can say exactly what a recursive delete removes.
- Extracted `listFolderDescendants()`, the descendants range-query previously inlined in `moveFolder()`, and reused it in both places.
- Recursive deletes record `folder` and `numDescendants` in the `asset.delete` action log.

**`ui/components/AssetBrowser/AssetBrowser.tsx`**

- `DeleteAssetModal` counts the contents of any selected folders when it opens (showing a small "Checking folder contents..." spinner, with the confirm button disabled until it resolves).
- Non-empty folders render a red warning — *"This folder is not empty. Deleting also permanently removes the 12 file(s) and 3 folder(s) nested inside."* — and the confirm button becomes **Delete all**. Empty folders warn about nothing and keep the plain **Delete** button.
- Selections above `MAX_FOLDER_LOOKUPS` (20 folders) skip the per-folder count to bound the query cost and fall back to a generic "everything nested inside" warning, as does a failed count.
- Dropped the now-inaccurate "Folders must be empty before they can be deleted." copy from the modal body and the delete-failure notification.

Underlying GCS files are still left untouched, since published docs may reference them.

## Testing

- New `ui/utils/assets.delete.test.ts` covers empty-folder deletes, the `FolderNotEmptyError`, recursive deletion across nesting levels, prefix-collision siblings (`marketing-archive` is not swept up by deleting `marketing`), write-batch splitting at 900 descendants, the action-log metadata, and `countFolderContents()`.
- `npx vitest run ui/` — 455 tests across 45 files pass.
- `eslint` and `prettier --check` pass on the changed files; `tsc --noEmit` reports only the pre-existing `Layout.tsx` import-assertion error.

---
_Generated by [Claude Code](https://claude.ai/code/session_019ZvSdTeFNKYC27yztDDky7)_

### DIFF
--- a/.changeset/olive-lions-repeat.md
+++ b/.changeset/olive-lions-repeat.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: warn and recursively delete non-empty asset folders
--- a/packages/root-cms/ui/components/AssetBrowser/AssetBrowser.css
+++ b/packages/root-cms/ui/components/AssetBrowser/AssetBrowser.css
@@ -334,6 +334,14 @@
   font-size: 14px;
 }
 
+.AssetBrowser__deleteModal__checking {
+  display: flex;
+  align-items: center;
+  gap: 8px;
+  font-size: 13px;
+  color: var(--color-text-gray);
+}
+
 .AssetBrowser__deleteModal__warning {
   font-size: 13px;
   color: #c92a2a;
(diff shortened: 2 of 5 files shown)


stevenle:
tested locally and this lgtm

PR #1380, opened 2026-08-20
feat: add <StyleDeps> for adding CSS deps of dynamic components

stevis:
## Problem
Root collects a route's CSS by walking its static Vite import graph — collectCss()
in build-asset-map.ts follows importedModules (manifest imports) and never
dynamicImports. Sites that resolve components dynamically (e.g. a CMS mapping
content-types → templates via import.meta.glob(..., { eager: true })) therefore
bundle every component's .module.scss into one shared chunk shipped on every page
— render-blocking and hurting LCP. Switching those to lazy import() drops the CSS
entirely, breaking styles.

## Solution
Add a first-class way to tell Root "this component rendered — include its CSS",
mirroring the existing collectElementDeps path used for custom elements.
- HtmlContext.componentDeps: string[]
- New <Deps src="..."> component (same shape as <Script>/<Head>), exported from core
- render.tsx resolves each componentDeps entry via assetMap.get(src).getCssDeps()
  and injects it (reusing the ?inline filter)

Reuses Root's own asset-map collector — no manifest re-parsing, no change to
collectCss traversal, fully backward-compatible and opt-in.

## Usage
```<Deps src="templates/Foo/Foo.tsx" />```

## Testing
- New build-asset-map.test.ts: a dynamically-imported component's CSS resolves by
  src while the route alone does not pull it in.
- vitest run src/render src/core → 29 pass. pnpm build passes. Pre-existing
  unrelated tsc/prettier issues on main are untouched.

## Changeset
Included — @blinkk/root minor.

### DIFF
--- a/.changeset/curly-jars-attend.md
+++ b/.changeset/curly-jars-attend.md
@@ -1,0 +1,11 @@
+---
+'@blinkk/root': minor
+---
+
+feat: add `<StyleDeps>` for per-page CSS of dynamically-imported components
+
+Root collects a route's CSS by walking its static import graph, so the
+`.module.scss` of a component loaded via `import()` is never linked. The new
+`<StyleDeps src="...">` component registers a rendered component's source path so
+Root resolves and injects its CSS deps, letting a page load only the CSS for the
+components it actually renders.
--- a/.changeset/curly-jars-attend.md
+++ b/.changeset/curly-jars-attend.md
@@ -1,11 +1,5 @@
 ---
-'@blinkk/root': minor
+'@blinkk/root': patch
 ---
 
-feat: add `<StyleDeps>` for per-page CSS of dynamically-imported components
-
-Root collects a route's CSS by walking its static import graph, so the
-`.module.scss` of a component loaded via `import()` is never linked. The new
-`<StyleDeps src="...">` component registers a rendered component's source path so
-Root resolves and injects its CSS deps, letting a page load only the CSS for the
-components it actually renders.
+feat: add <StyleDeps> for adding CSS deps of dynamic components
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by stevenle** [comment]
@stevis thanks for the PR. Was `<Deps>` the AI-proposed name? It feels too generic for what it does, I think maybe something with "Style" or "Stylesheet" or "CSS" in it would make it more clear, e.g. `<StyleDeps src={path}>`

**Comment by stevenle** [comment]
In case you need to test your changes today before making any changes upstream to root, there is a `useAsset(src)` hook that you can maybe leverage to fix:

```
/**
 * Returns an asset from the project's module graph. Paths are relative to the
 * project root and may optionally start with `/`.
 */
export function useAsset(src: string): Asset {
  const normalizedSrc = src.replace(/^\/+/, '');
  const asset = useAssetMap().get(normalizedSrc);
  if (!asset) {
    throw new Error(`asset not found: ${src}`);
  }
  return asset;
}
```

This will provide metadata from the vite dependency graph and may contain the dependent CSS files that you need.

stevis:
Yeah, this entire PR is AI-proposed :)

This isn't urgent, but I did notice this issue today doing some performance testing. I think the entire file gzip'd is small enough to not be an issue but it's obviously something that'll get worse over time.

Downstream, the idea is to include the needed module styles within Workspace's PageModule template, instead of the entire bundle. This seemed like the cleanest approach?

stevenle:
Got it, within a PageModules loop it might output the same StyleDeps tag multiple times if the same template is called over and over. Mind adding a test to verify that the style deps are de-duped? From the code I think it already is handled that way, but the test would ensure no regressions in the future.

stevenle (approved):

stevenle:
thanks for the PR!

PR #1388, opened 2026-08-24
feat: add paragraph size variants to the richtext editor

willthomson:
Paragraphs can now carry a size the same way headings carry a level. The size is stored as `data.size` on the paragraph block and rendered as `<p data-size="...">`.

Because the size lives *outside* `data.text`, it never touches the HTML tag allowlist, the unknown-tag parser, or the translation pipeline — translation extraction reads strings only, so a size can never orphan a translation.

## Sites declare their own sizes

```ts
schema.richtext({
  id: 'body',
  paragraphSizes: ['small', {value: 'large', label: 'Larger'}],
})
```

The values are the site's own — `small`, `large`, `body-3`, whatever the design system calls them — and each site decides in its CSS what they mean, exactly as each site already decides what `h2` means inside a body. Entries may be a bare string or `{value, label}` when the dropdown should read differently from the stored value.

Fields that omit `paragraphSizes` behave exactly as they do today, which is what makes this safe across every existing richtext field. There is no `'normal'` value — absence means the site's default body size, so no document needs migrating.

## Editor

A `SizedParagraphNode` (a `ParagraphNode` subclass) carries the size and mirrors it onto the editor DOM as `data-size`. Plain paragraphs keep using lexical's built-in `ParagraphNode`; the subclass is only created when an editor explicitly picks a size.

Both converters round-trip the size on **both** paths — root-level paragraphs and table-cell paragraphs — and they ship together, since root-cms uses a bespoke serialiser and parser rather than lexical's `exportDOM`. Pressing Enter carries the size to the next paragraph, matching Docs and Word.

The block-type dropdown gains one entry per declared size, under "Normal".

## Notes for reviewers

- **Lexical only, enforced.** `EnableEditorJSEditor` is a per-user preference on every project, and the legacy EditorJS editor knows nothing about paragraph size: `editor.save()` rebuilds every block from its own tools, so one edit there would silently strip `data.size` from paragraphs another user had sized. A field that offers sizes therefore always opens in the lexical editor, whatever the user's preference.
- **Sizes travel on the internal clipboard.** `PasteCleanupPlugin` returns early for `application/x-lexical-editor`, so copying between richtext fields carries the size regardless of what either field declares. This is deliberate: stored values are checked for usability, never against a field's list. A paragraph carrying an undeclared size shows the raw value in the dropdown rather than being mislabelled "Normal".
- **Missing site CSS is graceful.** `<p data-size="small">` with no CSS renders as an ordinary paragraph. The editor previews the two conventional values (`small`, `large`) so picking one is visible without the split preview; a size named anything else previews at normal size in the editor and is still correct on the site.
- **`__paragraphSize`, not `__size`.** `ElementNode.__size` is lexical's own child count — using it fails with `ElementNode.splice: start + deleteCount > oldSize`. Caught by the round-trip test.
- **`FloatingToolbarPlugin` is untouched.** It has no block-type dropdown, only inline formats, so there is nothing to keep in sync.
- The `docs/` sandbox template declares `paragraphSizes` and styles them, so the feature is exercisable in the local playground.

## Testing

New tests cover the converter round-trip at the root and inside tables, arbitrary site-defined values, unusable stored values degrading to a normal paragraph, Enter carrying the size forward, lexical serialize/reparse (the internal clipboard path), and the renderer's `data-size` output.

Not run locally: the Firestore-emulator suite (no JDK on this machine) and the visual suite — three richtext visual tests time out identically on a clean `main` after the recent Playwright bump.

Generated with Claude Code (Opus 5).

---

## Review history

Supersedes #1386, which was opened from a fork before I had write access here. Identical commits; that PR is closed. Three automated review threads were raised and addressed there:

- **Codex, P1 — legacy editor strips sizes.** `EditorJSEditor.save()` rebuilds every block from its own tools, so one edit by a user with `EnableEditorJSEditor` would have silently dropped `data.size`. Fixed in 9aeeb033: a field that offers sizes always opens in Lexical.
- **Copilot — the schema gate wasn't enough.** A field can hold sized paragraphs without declaring `paragraphSizes` (pasted in, and still rendered, because `p[data-size]` CSS is global rather than per field). Fixed in 76b8ac30: `testRichTextParagraphSizes()` also scans the stored value, including table cells, and the decision is latched.
- **Copilot — `insertNewAfter` should carry indent.** Declined. Lexical 0.49's own `ParagraphNode.insertNewAfter` never calls `setIndent`, and there is no `setIndent` in `LexicalSelection.ts`; a headless run confirms plain and sized paragraphs both reset indent to `0` on Enter. Adding it would make sized paragraphs the inconsistent ones. If carrying indent through Enter is wanted, it belongs in its own change applied to both.


stevenle:
thanks for the PR @willthomson . since the paragraph sizes are user-defined, a random/custom string wouldn't have css styles defined in the WYSIWYG lexical editor, which would mean it would default to the normal size (so the user won't have a visual indicator of the size difference).

i think we should maybe have some config setting that allows the developer to specify the css style(s) to apply for the custom sizes (and make sure to sanitize the user input so that there is no potential for code injection). and let's remove the following in `LexicalTheme.css`:

```
/*
 * Approximations for the two conventional paragraph sizes, so that picking one
 * is visible in the editor and not only in the split preview. Sites define
 * their own size values and their own CSS, so a size named anything else
 * renders at the normal size here and is still correct on the site.
 */
.LexicalTheme__paragraph[data-size='small'] {
  font-size: 0.875em;
}
.LexicalTheme__paragraph[data-size='large'] {
  font-size: 1.25em;
}
```

willthomson:
Thanks @stevenle — you and Will landed on this independently within a few minutes of each other, and both changes are already on the branch (they went up just after your comment, so you were reviewing `76b8ac30`).

**Hardcoded theme CSS removed** in `ac091a66`. That block was left over from when the size union really was `'small' | 'large'`; it should have gone when sizes became site-defined. Exactly the problem you describe — the toolbar would say "Tiny" over text that never changed size, which reads as a broken feature rather than a working one.

**Config setting** is `editorStyle`, per size, in `43fd086d`:

```ts
paragraphSizes: [
  {value: 'small', label: 'Small', editorStyle: {fontSize: '0.875em', lineHeight: 1.4}},
  {value: 'large', label: 'Larger', editorStyle: {fontSize: '1.25em', fontWeight: 500}},
]
```

Named `editorStyle` rather than `style` so the scope is legible at the call site — it never reaches the published page, which still carries `data-size` alone and is styled by the site's own CSS.

**On injection**, three layers:

1. Allowlist. Only `fontSize`, `lineHeight` and `fontWeight` survive normalization; everything else is dropped before it can reach the DOM. There's a test asserting `color` and `display` don't make it through. That also keeps the editor from being able to misrepresent the content (no `display: none`, no recoloring).
2. Values are applied through CSSOM property assignment (`dom.style.fontSize = value`), never string-interpolated into a stylesheet or a `style` attribute. A value like `1em; color: red` is invalid *for that property*, so the browser drops it — there's no way to break out into another declaration.
3. The values are developer-authored in `schema.ts`, not editor input from the DB, so this is defense in depth rather than the primary trust boundary.

Scope is deliberately typography only. Happy to widen it if a real case turns up — widening a type later is non-breaking, narrowing isn't.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

stevenle (approved):

PR #1390, opened 2026-08-25
chore: polish task details page layout and timeline density

stevenle:
UI polish for the task details page (`TaskPage`) — width, sticky sidebar offset, and timeline density.

### Wider page

The page's width caps are consolidated into a single `--task-page-max-width` custom property on `.TaskPage` and raised from 1120px to 1340px, applied to the header, the deleted banner, and the two-column body. The grid columns widen to match: main `760px → 960px`, sidebar `260–320px → 280–340px`. The inline title-edit form widens to 960px so it stays aligned with the main column.

Side effect worth noting: the header previously capped at 1040px, narrower than the 1120px body. It now shares the same cap as everything else, so the breadcrumbs/title row lines up with the content below it.

### Less whitespace above the sticky sidebar

`.TaskPage__side` used `top: calc(var(--header-height) + 20px)`. That offset assumed the window was the scroll container, but it isn't — `.Layout__main > *:first-child` gives `.TaskPage` itself `height: calc(100vh - 48px); overflow: auto`, so the scroll port already starts below the 48px CMS header.

There is a second contribution on top of that. Sticky offsets resolve against the scroll container's **content** box, so `.TaskPage`'s own `--page-padding-lg` (60px) stacks onto whatever `top` is set to. Measured in Chromium against a reproduction of this layout, the rendered gap between the CMS header and the pinned panel was:

| `top` | rendered gap |
| --- | --- |
| `calc(var(--header-height) + 20px)` (before) | 128px |
| `20px` | 80px |
| `-20px` (this PR) | 40px |

The negative value is deliberate and safe: the 60px of page padding absorbs it, so the panel is never clipped and keeps its rounded top corners. There's a comment on the rule explaining this, since a bare negative `top` reads like a typo otherwise.

### Deeplink offset

`scroll-margin-top` on `.TaskPage__timelineItem:target` (the `#comment-N` deeplink) had the same header double-count, so it drops from `calc(var(--header-height) + 16px)` to `16px`. Note that `scroll-margin` does *not* behave like the sticky offset above — it resolves against the padding box, so no padding compensation is needed here. Verified both cases in Chromium: the target now lands 16px below the top of the scroll port instead of 64px.

### Tighter timeline

Timeline markers shrink from 32px to 24px, with the Tabler glyphs inside going from `size={15}` to `size={13}`. Surrounding type and spacing come in with them:

| | before | after |
| --- | --- | --- |
| marker | 32px | 24px |
| column gutter | 12px | 10px |
| event text | 13px / 1.5 | 12px / 1.45 |
| value chip | 12px, `2px 5px` | 11px, `1px 5px` |
| row padding-bottom | 18px | 12px |
| reply indent / gap | 42px / 12px | 34px / 10px |

Three values are derived from the marker size rather than picked, and were re-measured after the change: the rail's `left` (11px, so its 2px width centers on the 24px marker), the `first-child`/`last-child` rail stubs, and the event text's `padding-top` (4px, which puts the first line's center exactly on the marker's center). A rendered check confirms rail center and marker center both land at 12px, and the text/marker vertical delta is 0px. Net effect is event rows going from ~43.5px to ~32.4px tall.

Comment card internals (14px body, 13px header) are left alone — only the timeline row scaffolding changed.

### Testing

Layout and timeline geometry were verified by rendering the real `TaskPage.css` in headless Chromium against markup mirroring the component's DOM, and measuring marker/rail/text positions and the pinned panel and deeplink offsets. Prettier passes on both changed files. `eslint` couldn't run in this environment (monorepo deps aren't installed) and doesn't cover CSS. There are no visual test goldens for `TaskPage`, so nothing needed regenerating.

### DIFF
--- a/packages/root-cms/ui/pages/TaskPage/TaskPage.css
+++ b/packages/root-cms/ui/pages/TaskPage/TaskPage.css
@@ -1,11 +1,13 @@
 .TaskPage {
+  /* Shared width cap for the page's header, banners, and two-column body. */
+  --task-page-max-width: 1260px;
   background: var(--page-background);
   padding: var(--page-padding-lg);
 }
 
 .TaskPage__header {
   margin-bottom: 28px;
-  max-width: 1040px;
+  max-width: var(--task-page-max-width);
 }
 
 .TaskPage__header__breadcrumbs {
@@ -41,7 +43,7 @@
   display: grid;
   gap: 8px;
   grid-template-columns: minmax(220px, 1fr) auto auto;
-  max-width: 760px;
+  max-width: 880px;
   min-width: 0;
   width: 100%;
 }
@@ -107,8 +109,8 @@
   align-items: start;
   display: grid;
   gap: 28px;
-  grid-template-columns: minmax(0, 760px) minmax(260px, 320px);
-  max-width: 1120px;
+  grid-template-columns: minmax(0, 880px) minmax(280px, 340px);
+  max-width: var(--task-page-max-width);
 }
 
 .TaskPage__main {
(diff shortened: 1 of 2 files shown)


stevenle:
Screenshot 2026-08-24 at 5 38 22 PM

PR #1392, opened 2026-08-25
feat: add user autocomplete to the task assignee and cc fields

stevenle:
Replaces the plain email text inputs on the task details page with autocomplete pickers sourced from the project's user list, so assigning or cc'ing a teammate is a matter of searching for their name rather than typing their address exactly.

## Changes

**`UserSelect` / `UserMultiSelect`** (`ui/components/UserSelect/`)

- Each dropdown option renders the user's avatar (profile photo, or the initials fallback when there's none), their display name, and their email.
- Search matches against both the display name and the email, so `lovelace` and `ada@` both find Ada.
- Any address that isn't a project user can still be entered by hand — typing a valid email offers a `+ Use "…"` / `+ Add "…"` option. Non-email values don't offer it.
- `UserMultiSelect` renders each selection as a removable chip with an avatar.

**`useProjectUsers()`** (`ui/hooks/useProjectUsers.ts`)

Merges the project's roles map with the saved user profiles (which supply display names and photos), dropping domain wildcards like `*@example.com` since they aren't addressable people. Both sources are already cached in memory; `listUserProfiles()` now caches the list itself (5 min TTL, with in-flight de-duping), so the pickers open instantly after the first load instead of re-reading the collection on every mount.

**Task details page** (`ui/pages/TaskPage/TaskPage.tsx`)

Assignee uses `UserSelect`, cc uses `UserMultiSelect`. The separate cc chip list below the input is gone — chips now live in the field itself — but the "copy full email addresses instead of the short names" behavior is preserved: chips carry a `data-user-email` attribute that the existing `onCopy` handler reads.

## Behavior change worth flagging

Cc'ing someone who isn't a member of the project used to be rejected with a "not a member of this project" notification. That check is removed, since free-value entry was explicitly asked for and the assignee field never enforced membership either. Values still have to look like email addresses.

## Testing

- `user-select-data.test.ts` covers the extracted logic: email validation, option building (project users + typed-in values + unknown selections, de-duped), name/email matching, and when the "add this address" option is offered. 14 tests, passing.
- Verified end-to-end in Chromium (throwaway harness, not committed): options render with avatars and emails, filtering by email prefix narrows the list, selecting updates the value, and adding a free-value email to cc produces the expected chip.
- `eslint` clean on all touched files; `tsc --noEmit` shows no new errors; the UI bundle builds.

Mantine's jsdom rendering doesn't work in this repo's unit suite (existing component tests mock `@mantine/core` wholesale), which is why the component logic was extracted into a plain module rather than tested through the rendered component.

## Not included

The "new task" form in `TaskManager` still uses a plain text input for assignee — left alone since the request was scoped to the details page, but it's the obvious follow-up.

https://claude.ai/code/session_01NdfFotui2uNUiuaTD28WCs

---
_Generated by [Claude Code](https://claude.ai/code/session_01NdfFotui2uNUiuaTD28WCs)_

### DIFF
--- a/.changeset/shaggy-tasks-autocomplete.md
+++ b/.changeset/shaggy-tasks-autocomplete.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: add user autocomplete to the task assignee and cc fields
--- a/packages/root-cms/ui/components/UserSelect/UserSelect.css
+++ b/packages/root-cms/ui/components/UserSelect/UserSelect.css
@@ -1,0 +1,58 @@
+.UserSelect .mantine-Select-input,
+.UserSelect .mantine-MultiSelect-input {
+  min-height: 30px;
+}
+
+.UserSelect__item {
+  align-items: center;
+  display: flex;
+  gap: 8px;
+  min-width: 0;
+}
+
+.UserSelect__item__text {
+  display: flex;
+  flex-direction: column;
+  line-height: 1.3;
+  min-width: 0;
+}
+
+.UserSelect__item__label {
+  font-size: 12px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
+.UserSelect__item__email {
+  color: var(--color-text-gray);
+  font-size: 11px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
+.UserSelect__value {
+  align-items: center;
+  background: #f6f7f8;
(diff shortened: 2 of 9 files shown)


stevenle:
@jeremydw i tested this, and i'm not super in love with it (yet). i'm considering just doing what github does. thoughts?

Screenshot 2026-08-24 at 8 42 55 PM

jeremydw:
It could maybe be polished a bit but IMO it LGTM for the initial version. After we put tasks through more power user use cases and big projects I'm sure we can come back and refine things further!

PR #1394, opened 2026-08-25
fix: flush rich text editors when draft edits are discarded

### LINKED ISSUE #1393: "Discard draft edits" doesn't flush rich text editors
Steps to reproduce:

1. Open a published doc
2. Make changes to a rich text field in draft
3. Observe changes
4. Use "discard draft edits"
5. The rich text field _should_ be flushed back to the published state, but it remains as-is.

stevenle:
Fixes #1393.

## Problem

Both rich text editors ignored an incoming `value` unless its `time` was newer than the value they last rendered:

- `lexical/plugins/OnChangePlugin.tsx` — `if (time > timeSavedRef.current)`
- `editorjs/EditorJSEditor.tsx` — `if (newValueTime > currentTime && ...)`

"Discard draft edits" copies the published doc over the draft, and the published rich text data was serialized *before* the draft edits, so its `time` is older. The draft doc's snapshot listener pushed the published value down to the field, the editor compared timestamps, and kept showing the discarded content. The same gate drops any external replacement that carries an older timestamp — version restore, and undo/redo (which by construction writes the *previous* value).

## Fix

Compare the incoming value's `blocks` against the content the editor already holds, instead of comparing timestamps:

- New `testSameRichTextContent()` in `shared/richtext.ts` compares blocks only (`time`/`version` are editor bookkeeping) and treats missing/`null`/`undefined` keys as equivalent, since values that round-trip through Firestore lose their `undefined` keys.
- Both editors track the content currently reflected in the editor and re-render only when an incoming value differs from it. Echoes of the editor's own change still no-op, so typing never resets the cursor; a genuine replacement re-renders whatever its timestamp says.
- `RichTextField` also compares content rather than `time` before writing to the draft, so an editor re-emitting the value it was just handed (with a fresh `time`) doesn't immediately re-dirty the doc that was just discarded.
- Emptying a field now clears the editor too — previously an empty value failed the `blocks.length > 0` validity check in EditorJS and was dropped.
- Removed `validateRichTextData()`, which had no remaining callers.

## Why a content compare rather than a cheaper timestamp check

Two cheaper-looking alternatives don't cover the bug:

- **`time !== lastTime` instead of `>`** would re-render on the discard, but `time` is a serialization timestamp, not a content identity. After the discard the editor re-serializes and emits with a fresh `Date.now()`; a timestamp comparison in `RichTextField` reads that echo as a change and writes it straight back to the draft, so the doc goes "saving..." and is modified again seconds after the user discarded. It also re-renders (resetting the cursor) whenever a content-identical value arrives with a new timestamp.
- **Resetting the editor from the discard action** (remount / key bump) only covers the button. `cmsRevertDraft()` is also reachable from the collection list page and the `doc_revertDraft` AI tool, and a discard performed in another tab reaches this editor only through the snapshot listener — the same path version restore and undo/redo use.

The cost is not where it looks. Measured on a synthetic doc (numbers are per call, Node 22, this container):

| | 65 blocks (10KB) | 260 blocks (40KB) | 1300 blocks (200KB) |
|---|---|---|---|
| same-reference (the typing path) | 0.05µs | 0.04µs | 0.04µs |
| deep walk, equal | 24µs | 58µs | 308µs |
| `deepEqual(blocks)` the trie store already runs per update | 9µs | 36µs | 185µs |

The keystroke path exits on reference equality, because an editor's own value object is what comes back to it — the deep walk only runs for values that arrived from somewhere else (db, undo, revert), i.e. at human rate. Meanwhile `JsonTrieStore.update()` already runs a full `deepEqual` over the same blocks on *every* `updateKey()`, so a keystroke pays ~4600× more there than in this check.

The principled way to delete the heuristic entirely is to tag draft updates with an origin so a field can recognize its own writes and skip the comparison altogether — but that changes the controller/store subscription API every field uses, so it's left out of this fix.

## Tests

- `shared/richtext.test.ts` — unit tests for `testSameRichTextContent()`.
- `lexical/plugins/OnChangePlugin.test.tsx` (new) — mounts a minimal lexical composer around the plugin and covers: rendering an external replacement with an older timestamp (verified to fail before the fix), clearing on an emptied value, and ignoring echoes/timestamp-only changes (asserted via editor state identity, i.e. no re-render).
- `DocEditor/fields/RichTextField.test.tsx` (new) — covers passing db values through and not saving an editor echo of the current value.

`pnpm --filter @blinkk/root-cms vitest run ui shared` → 69 files / 740 tests pass; `tsc --noEmit` and eslint clean on the changed files.

## Not covered

The EditorJS path (legacy, opt-in via the `EnableEditorJSEditor` preference) got the same fix but no automated test — EditorJS doesn't mount usefully under jsdom.

### DIFF
--- a/.changeset/lucky-moons-tickle.md
+++ b/.changeset/lucky-moons-tickle.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+fix: flush rich text editors when draft edits are discarded
--- a/packages/root-cms/shared/richtext.test.ts
+++ b/packages/root-cms/shared/richtext.test.ts
@@ -3,6 +3,8 @@ import {
   getRichTextParagraphEditorStyles,
   normalizeRichTextParagraphSizes,
   parseRichTextParagraphSize,
+  RichTextData,
+  testSameRichTextContent,
 } from './richtext.js';
 
 describe('parseRichTextParagraphSize', () => {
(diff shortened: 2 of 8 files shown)


stevenle:
@jeremydw mind running a quick test on this based on your steps to replicate the issue?

PR #1408, opened 2026-09-01
fix: scroll long components dropdown in the rich text toolbar

### LINKED ISSUE #1407: Improve UI when too many rich text block components cause overscroll
In Rich Text fields, when we configure many block components, the dropdown needs some work (see screenshot; note grey areas are redacted, not a rendering bug).

Image

1. Icons shrink (should maintain consistent size)
2. Long block names overflow (should truncate with ellipsis)
3. Dropdown is too long (should scroll vertically)

Additionally, the custom block component icon is currently an empty square; this makes it look sort of like an unchecked checkbox. We should consider changing it to a different icon like tabler's "components" https://tabler.io/icons?icon=components or cube https://tabler.io/icons?icon=cube or puzzle piece https://tabler.io/icons?icon=puzzle

stevenle:
Fixes #1407 three issues with the **Components** dropdown in the rich text (Lexical) toolbar.

### 1. Vertical overflow with long component lists

Mantine's `Menu` body has no max height, so a project with many custom components rendered a menu that ran straight off the bottom of the viewport with no way to reach the items below the fold.

The menu body now gets `max-height: min(60vh, 480px)` plus `overflow-y: auto` (scoped via `classNames={{body}}`, since the menu renders in a portal).

Measured in Chromium at a 900×700 viewport with 27 components: menu height **1641px → 378px** with `scrollHeight` 2185, i.e. it now scrolls inside the menu.

### 2. Icon squished by long component names

`.mantine-Menu-itemIcon` is a flex item with the default `flex-shrink: 1`, and the sibling `itemBody` (`flex: 1`) won't shrink below its min-content width. A component label too long for one line therefore pushed the row wider than the menu and the icon absorbed the overflow.

For a name without spaces this collapsed the icon entirely — measured **0.00×0.00px** before the fix. The label now wraps with `overflow-wrap: anywhere` and the icon is pinned with `flex-shrink: 0`; same case measures **14.40×14.40px** after. `line-height` on the label is bumped from Mantine's `1` to `1.3` so wrapped names aren't cramped.

### 3. Default component icon

The fallback icon for custom block components was `IconSquare`, which reads as an unchecked checkbox. Switched to `IconCube`.

### Testing

- `pnpm build` passes.
- `pnpm lint` reports no new problems in the changed files (the repo has pre-existing lint errors elsewhere, all in `packages/root`).
- Verified in Chromium via a throwaway visual test that rendered the editor with 27 block components (long spaced names, one long unbroken name) and one long inline component name, measuring the menu body and every item icon before and after. The scratch test was removed before committing; no committed screenshot goldens change, since none of them capture an open Components menu.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TdbXifpmvTejSCdKRtTy29)_

### DIFF
--- a/.changeset/rich-text-components-dropdown.md
+++ b/.changeset/rich-text-components-dropdown.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+fix: scroll long components dropdown in the rich text toolbar
--- a/packages/root-cms/ui/components/RichTextEditor/lexical/LexicalEditor.css
+++ b/packages/root-cms/ui/components/RichTextEditor/lexical/LexicalEditor.css
@@ -77,24 +77,34 @@
   font-size: 12px;
 }
 
-/* Constrain the height of the components dropdown so that long lists of custom
+/* Widen the components dropdown past Mantine's default 200px so that most
+ * component names fit, and constrain its height so that long lists of custom
  * components scroll within the menu instead of overflowing the viewport. */
 .LexicalEditor__toolbar__insertDropdown__body {
   max-height: min(60vh, 480px);
   overflow-x: hidden;
   overflow-y: auto;
+  width: 280px;
 }
 
-/* Component names that don't fit on a single line push the flex row wider than
- * the menu, which squishes (and can fully collapse) the item's icon. Let the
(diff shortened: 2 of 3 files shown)


stevenle:
@jeremydw can you test this on the project in question?

PR #1409, opened 2026-09-01
feat: add a search filter to the rich text components dropdown

stevenle:
Adds type-to-filter to the **Components** dropdown in the rich text toolbar, so projects with a lot of custom components don't have to scroll a long menu. Builds on #1408.

### Behavior

A search input appears above the list once there are at least 8 components (`COMPONENT_FILTER_MIN_COMPONENTS`), so small projects keep the plain menu. It filters every entry — inline components, block components, and the built-in Table / Horizontal Rule — matching on both the display label and the schema `name`, and section headings disappear along with their contents.

Keyboard:

| Key | Action |
| --- | --- |
| *(on open)* | the input is focused automatically |
| ArrowDown | moves into the list; Mantine's own arrow navigation takes over from there |
| Enter | inserts the first match |
| Escape | clears the query; a second Escape closes the menu |

The query resets whenever the menu is reopened, and an empty result shows "No components found."

### Implementation notes

- The input is focused via a **callback ref**, not an effect. The popper mounts the dropdown body asynchronously, so an effect keyed on the menu's open state runs before the input exists — confirmed empirically (`document.activeElement` was still the toolbar button).
- Mantine 4 tracks the highlighted menu item through `mouseenter`, not focus, and its items set `outline: 0`. Focusing an item directly would therefore move focus invisibly, so `focusMenuItem()` dispatches a `mouseenter` alongside `.focus()`. That also leaves Mantine's internal `hovered` index consistent, so its ArrowDown/ArrowUp handling continues correctly from the item you landed on rather than restarting at the top.
- The menu is now controlled (`opened`/`onOpen`/`onClose`) so Escape can close it — Mantine's own key handling is bound to the control and to the items, and never sees keystrokes typed in the input.

### Testing

- New `ToolbarPlugin.visual.test.tsx` — 8 browser tests covering the threshold, autofocus, filtering by label and by name, the empty state, Enter-inserts-first-match (asserted through the real HTML block modal opening), ArrowDown, Escape, and reset-on-reopen. No screenshot assertions, so they are platform-independent. Ran repeatedly in Chromium: 8/8 every time.
- `pnpm test` (root-cms, Firestore emulator): 1320 passed.
- `pnpm lint` and `pnpm build` clean; `tsc --noEmit` reports nothing in the changed files.

**Unrelated pre-existing flake, worth flagging:** `LexicalEditor.visual.test.tsx > renders html block` times out intermittently in my environment — it failed 4 of 6 runs on a clean checkout with no changes at all. I first suspected this change and bisected it; it reproduces without any of this work, and `LexicalEditor.visual.test.tsx` is untouched here. Also note that CI installs Playwright but `pnpm test` excludes `*.visual.test.tsx`, so the visual suite — including these new tests — only runs via `pnpm test:visual`.

https://claude.ai/code/session_01TdbXifpmvTejSCdKRtTy29

---
_Generated by [Claude Code](https://claude.ai/code/session_01TdbXifpmvTejSCdKRtTy29)_

### DIFF
--- a/.changeset/richtext-components-search.md
+++ b/.changeset/richtext-components-search.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: add a search filter to the rich text components dropdown
--- a/packages/root-cms/ui/components/RichTextEditor/lexical/LexicalEditor.css
+++ b/packages/root-cms/ui/components/RichTextEditor/lexical/LexicalEditor.css
@@ -107,6 +107,24 @@
   white-space: nowrap;
 }
 
+/* Filter input pinned to the top of the components dropdown. The negative
+ * margins let it span the menu's own padding. */
+.LexicalEditor__toolbar__insertDropdown__filter {
+  background: #fff;
+  border-bottom: 1px solid #e9ecef;
+  margin: -4px -4px 4px;
+  padding: 6px;
+  position: sticky;
+  top: -4px;
+  z-index: 1;
+}
+
+.LexicalEditor__toolbar__insertDropdown__empty {
+  color: #868e96;
+  font-size: 12px;
+  padding: 8px 12px;
+}
+
 .LexicalEditor__editor {
   border: 1px solid #ced4da;
   background: #fff;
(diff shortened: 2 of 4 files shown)


stevenle:
@jeremydw mind testing this one on your project? note: the filter searchbox only appears for > 8 components.

jeremydw:
Do you think we should enable it regardless? That way users can type in and use the keyboard more effectively even if the list isn't long

stevenle:
i think there should be some practical min, even if it's a small number. e.g. 2-3 components i think a searchbox would be overkill. so i can reduce the min to ~4 if you think 8 is too large.

PR #1414, opened 2026-09-03
feat: theme stylesheet hook (experimental)

willthomson:
## What

Themes for the CMS UI — **experimental**: the markup changes as features are added, so a theme may need adjusting from version to version. The `--cms-*` custom properties in `ui/styles/theme.css` are the part meant to last.

Themes come from the project. `cmsPlugin` takes the list, so a package or a plugin can export one, and picks which applies until a user chooses:

```ts
cmsPlugin({
  themes: [
    {id: 'clarity', name: 'Clarity', file: './cms/clarity.css'},
    ...somePlugin.themes,
  ],
  defaultTheme: 'clarity',
})
```

Every user picks their own theme from that list under Settings → User Preferences, or the stock look. CSS can also be added from the CMS without a deploy — site-wide under Site Settings, per user under User Preferences — pasted directly or via `@import url("https://…")` of a hosted file. Cascade: base, theme, site CSS, user CSS. Nothing changes for a project that registers no themes.

No themes ship with root-cms: the mechanism and the token contract are the change. What a themed CMS should look like can be worked out in projects first.

## Why

Projects have had no way to restyle the CMS: `cmsPlugin` had no CSS hook and the page is rendered server-side. The immediate need was making nested field groups look nested — a group's fields render on the same edge as the fields around them, so nesting is invisible unless you read the labels — and the built-in `drawerOptions: {inline: true}` trades the group header's emphasis for its indent. A theme can have both, and the mechanism is general.

## How

- `ui/styles/theme.css` holds every `--cms-*` custom property at its stock value, each with a note on what it does. It is the contract and the file to copy to start a theme. The properties are used without per-use fallbacks, so there is one place the stock values live. The Mantine overrides that file used to hold are now `ui/styles/mantine.css`.
- `core/theme.ts` reads a theme — `file` relative to the project root, then `css` — and hashes it. A bad id or a missing file is warned about once and skipped rather than taking the CMS down.
- `core/plugin.ts` serves each at `/cms/themes/<id>.css` behind the CMS's login, with `Cache-Control: immutable` when the request carries the current content hash and `no-cache` otherwise. Files are read on each request, so an edit shows on the next reload.
- `core/app.tsx` links the default as `<link id="root-cms-theme">` after `ui.css` and passes the registered themes to the client.
- `ui/hooks/useTheme.tsx` (`ThemeLoader`) points that link at the user's choice, and keeps two `<style>` elements in sync with the site's and the user's custom CSS. Settings gets a theme picker (user) and "Custom CSS" textareas (site: admins; user: self), saved on blur.

## What is supported

The `--cms-*` custom properties are the contract — names and meanings kept stable across releases, all listed in `ui/styles/theme.css` and explained in `packages/root-cms/docs/themes.md`. Class-name selectors are documented as unsupported: they work today but the markup may change with any release.

## Testing

- `pnpm build` for root-cms; `tsc --noEmit -p packages/root-cms/tsconfig.build.json` clean; 13 unit tests for the loader, id validation and default resolution (`core/theme.test.ts`).
- Exercised in a real CMS with two themes registered and one set as default: both served with the expected cache headers, an unknown id and a path-traversal attempt both 404, the picker lists project default / stock / each theme, switching swaps the stylesheet live and the choice survives a reload, and an edit to a theme file shows on the next reload.
- Confirmed a project with no themes is unchanged: every property falls back to the stock value in `theme.css`.
- Not run locally: the emulator-backed suite (CI covers it) and `test:visual`.


stevenle:
if i'm reading this right, this seems like it would set the theme on cms for ALL users of the cms, which would prevent them from having choice.

imo if we do wish to support themes, we should allow users to choose which theme they prefer with an optional default for the project.

theme'ing in general is a tricky because we may eventually change our implementation from version to version, which means different dom structures and different elements. if we add theme support that allows users to provide their own css, we either need to provide an api or contract to users for "these css selectors will always work" or users will need to understand that things may change on them upstream which will break their theme.

stevenle:
in jellyfin, there is a theme customization option which allows users to provide their own css, someone could copy/paste it directly from a github repo for example, or use an `@import` css declaration to pull from a hosted css file.

i think maybe something like that would work for now (somewhere on the admin page), and we make it clear that this theme api is experimental and may not work from version-to-version as we continually add new features.

willthomson:
Thanks for the review, Steven. Both points landed. A project-wide theme with no way out isn't fair on the people in the editor every day, and anything keyed to the editor's DOM is going to drift as the UI moves. Reworked accordingly.

The built-in theme is gone. What's left is the hook: `cmsPlugin({theme: {file: './cms-theme.css'}})` points at a stylesheet in the project, served at `/cms/theme.css` behind the editor login and linked after `ui.css`, nothing inlined. Each user gets a switch under User Preferences to turn the project theme off, and there are Custom CSS fields at site and user level (the Jellyfin approach, `@import url(...)` works), so none of it needs a deploy. The `--cms-*` custom properties are documented as the stable surface, class-name selectors are called out as unsupported, and it's marked experimental in the README, the option's JSDoc and the changeset.

I'll work out what a themed editor should actually look like on a project first and come back with something concrete if it earns its place. Ready for another look when you have a moment.

.changeset/cms-custom-css.md
```diff
@@ -0,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: editor theme CSS (experimental). `cmsPlugin({theme: {file}})` points at a stylesheet that styles the document editor for everyone on the project (served behind the editor's login, re-read on each request); each user can switch it off for themselves under Settings → User Preferences. CSS can also be added from the CMS without a deploy: site-wide under Site Settings and per user under User Preferences, including `@import` of a hosted file. The editor stylesheet exposes `--cms-*` custom properties with stock defaults as the surface meant to last (see `docs/editor-theme.md`); the editor's markup may change between versions.
```
stevenle: update changeset (should use the same format as the PR title)

packages/root-cms/docs/editor-theme.md
```diff
@@ -0,0 +1,92 @@
+# Editor theme
```
stevenle: "Editor theme" doesn't feel inclusive enough, we want to allow the user to be able to customize ALL parts of the CMS (not just the editor).

packages/root-cms/core/editor-theme.ts
```diff
@@ -0,0 +1,67 @@
...
+  css?: string;
+}
+
+/**
+ * Where the project theme is served, behind the same login as the editor
+ * page that links it.
+ */
+export const THEME_URL = '/cms/theme.css';
+
+export interface EditorTheme {
```
stevenle: Themes don't apply only to the editor, it should apply to the entire CMS.

packages/root-cms/ui/components/DocEditor/DocEditor.css
```diff
@@ -1,5 +1,15 @@
+/*
+ * Type and rhythm dials for themes (see the `theme` plugin option). The
+ * defaults reproduce the stock editor exactly.
+ *
+ *   --cms-font-size   base size for the editor's own text
```
stevenle: Instead of documenting in a comment like this, we should have a single theme.css file that a user would be able to copy the entire file from to apply their own customizations on top. That would allow the file to be self-documenting in terms of what customizations we support.

packages/root-cms/core/plugin.ts
```diff
@@ -438,6 +440,36 @@ export type CMSPluginOptions = {
...
+   * without a deploy: for the whole project under Settings → Site Settings,
+   * and per user under Settings → User Preferences — including an
+   * `@import url(…)` of a hosted file.
+   *
+   * The supported surface is the `--cms-*` custom properties, which are
+   * kept stable across versions. Rules that target the editor's class names
+   * or DOM are unsupported: they work today but may break with any release,
+   * since the editor's markup is free to change.
+   */
+  theme?: CMSTheme;
```
stevenle: Remove this, I think plugins should define themes, and the project should be able to accept multiple theme options. We can accept a "default theme" configuration here though which would apply if the user hasn't selected a theme.

packages/root-cms/ui/ui.tsx
```diff
@@ -344,6 +347,7 @@ function App() {
             <SiteSettingsProvider>
               <PendingReleasesProvider>
                 <UserPreferencesProvider>
+                  <EditorThemeApplier />
```
stevenle: "theme applier" seems like a weird name, can you suggest a few alternatives?

.changeset/cms-custom-css.md
willthomson: Done — changeset now reads `feat: theme stylesheet hook (experimental). …`, matching the PR title.

packages/root-cms/docs/editor-theme.md
willthomson: Agreed. Renamed throughout: `docs/themes.md`, `core/theme.ts`, `useTheme.tsx`, and the option is `themes`. Nothing in the wording ties it to the editor now — the stylesheet is loaded page-wide, so it always could style all of the CMS.

packages/root-cms/core/editor-theme.ts
willthomson: Agreed, and the naming was the only thing saying otherwise. `CMSTheme` now describes a theme for the CMS UI; the interface picked up `id` and `name` for the picker too.

packages/root-cms/ui/components/DocEditor/DocEditor.css
willthomson: Good call — that's much better than a comment nobody sees. `ui/styles/theme.css` now holds every `--cms-*` property at its stock value with a line on what each does, and it's the file to copy to start a theme. The per-use fallbacks in DocEditor.css are gone with it, so the stock values live in exactly one place. (The Mantine overrides that file used to hold moved to `mantine.css`.)

packages/root-cms/core/plugin.ts
willthomson: Done. `themes` takes a list, so a plugin or package can export one and the project spreads it in, and `defaultTheme` is the default-only config you describe — it just names an id from the list. Users pick from the same list under Settings → User Preferences. Nothing ships with root-cms.

packages/root-cms/ui/ui.tsx
willthomson: It was doing two jobs and named after neither. It's `ThemeLoader` now. Others I considered: `ThemeStyles`, `ThemeStylesheets`, `ThemeProvider` (misleading — no context), `ThemeEffects`. Happy to take another if you prefer one.

.changeset/cms-custom-css.md
stevenle: @willthomson i know ai is automating a lot of this PR for you but it'd be great if you could verify the changes it makes and that it's aligned with the feedback. this changeset still needs to be updated.

willthomson: Apologies, that one was an oversight on my part. Fixed in 5c823e8: single line now, same as the others. Point taken on checking the changes properly before replying.

willthomson: Apologies, that one was an oversight on my part. Fixed now: single line, same as the others. Point taken on checking the changes properly before replying.

stevenle (approved):

PR #1416, opened 2026-09-03
feat: add follow mode to the document preview

willthomson:
Adds a "follow mode" toggle to the document editor header. When it's on, navigating within the preview pane switches the editor to the corresponding CMS document, so clicking a link in the preview brings the fields along with it.

> [!NOTE]
> Stacked on #1415, which fixes a pre-existing bug that follow mode would otherwise hit on every use. Review that one first; this PR targets its branch and should be rebased onto `main` once it lands.

## How a url is matched to a document

The forward mapping in `doc-urls.ts` is lossy — a slug's `--` separators become slashes, `/index` collapses to `/`, repeated slashes collapse — so this can't be a true inverse. `ui/utils/preview-doc-match.ts` instead extracts slugs loosely and then **verifies each candidate by running the real forward functions** (`getDocPreviewPath` / `getDocServingPath`) and requiring an exact match. That keeps the matcher bug-compatible with the forward path by construction, so `--`, `/index`, base paths, locale url formats and trailing-slash normalization need no special handling.

The result is a *ranked list*, not a single document, because a catch-all `/[...slug]` collection matches every url. In the repo's own `docs/` project, `/blog/sandbox/foo` legitimately matches three collections. Candidates are ranked by literal-prefix specificity, then captured-segment count, then collection id.

Ranking alone isn't enough, so `ui/utils/follow-mode.ts` confirms the document **actually exists** before routing: `DraftDocController` loads a missing doc as an empty editor rather than erroring, so following an unverified candidate would silently open a blank document. Candidates are probed in rank order, first hit wins, and misses are memoized for 60s (`getDocFromCacheOrFetch` only caches hits, so a page that isn't a document would otherwise re-read every candidate on every load).

## Timing

Both panes swap in the same frame:

- **On click**, the resolve and existence probe start immediately, in parallel with the page load. This also warms the db cache so the document is ready.
- **On the iframe's `load` event**, the resolved target is committed and the editor routes.

Resolving on the `load` event alone made the editor visibly lag the preview; committing as soon as the lookup finished made it arrive *before* the preview. Splitting resolution from the commit aligns them. Navigations with no click behind them — form posts, redirects, back/forward — resolve and commit together at `load`.

Switching follow mode on also reconciles: if you navigated with it off, the two panes are out of sync, so enabling it catches the editor up to whatever the preview is showing.

## Keeping the preview in draft mode

`preview=true` is what selects draft mode (`core/route.ts:610`), and a site's own links have no reason to carry a CMS-only param. Without handling, following a link would put the editor on a document's *draft* while the pane beside it rendered that document's *published* page.

A bubble-phase click listener on the preview document rewrites same-origin link navigations to keep `preview=true`. Rewriting the click (rather than re-navigating afterwards) is what keeps this to a single page load. Bubble phase so a previewed site that handles its own link clicks still wins; in-page hash links are skipped so a scroll doesn't become a page load. Form posts and JS redirects still land in published mode until the next reload.

## Safety

- **Unsaved changes.** Autosave runs on a 3s delay, so a click can land mid-window. The pending update is flushed before the route changes, and the controller flushes again when disposed.
- **Loops.** Following is skipped when any candidate is the document already being edited, and a sliding-window rate limit (5 follows in 5s) disarms the mode with a notification if a page bounces the preview between documents.
- **Staleness.** An in-flight follow is discarded if a newer navigation supersedes it, if follow mode is switched off, or if the preview unmounts.
- **History.** `route(url, replace)` — the iframe's own navigation already added a session history entry, so pushing another would make Back take two presses per followed link.
- Pages that aren't documents (list pages, 404s, external routes) are silently ignored; in follow mode they're common enough that notifying about each would be unusable.

## Testing

30 new unit tests covering the url matcher and the resolver (`427 passed` across the ui utils suite). Both new modules are pure enough to test without the Firestore emulator.

Manually exercised against a real site with live Firebase credentials, since the `docs/` project has no `.env`.

### DIFF
--- a/.changeset/follow-mode.md
+++ b/.changeset/follow-mode.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+feat: add follow mode to the document preview
--- a/packages/root-cms/ui/pages/DocumentPage/DocumentPage.css
+++ b/packages/root-cms/ui/pages/DocumentPage/DocumentPage.css
@@ -312,3 +312,8 @@
   overflow: hidden;
   background: white;
 }
+
+.DocumentPage__side__header__followToggle.active {
+  color: var(--mantine-color-blue-7, #1c7ed6);
+  background-color: var(--mantine-color-blue-0, #e7f5ff);
+}
--- a/packages/root-cms/ui/pages/DocumentPage/DocumentPage.tsx
+++ b/packages/root-cms/ui/pages/DocumentPage/DocumentPage.tsx
@@ -1,7 +1,8 @@
 import './DocumentPage.css';
 
-import {ActionIcon, Button, Tooltip} from '@mantine/core';
+import {ActionIcon, Button, Loader, Tooltip} from '@mantine/core';
 import {useHotkeys} from '@mantine/hooks';
+import {showNotification} from '@mantine/notifications';
 import {
   IconArrowLeft,
   IconArrowUpRight,
@@ -9,8 +10,10 @@ import {
   IconDeviceFloppy,
   IconLayoutSidebarRightCollapse,
   IconLayoutSidebarRightExpand,
+  IconRoute,
 } from '@tabler/icons-preact';
 import {useCallback, useEffect, useMemo, useRef, useState} from 'preact/hooks';
+import {useLocation} from 'preact-iso';
 import {ChecksPanel} from '../../components/ChecksPanel/ChecksPanel.js';
 import {ConditionalTooltip} from '../../components/ConditionalTooltip/ConditionalTooltip.js';
 import {DocEditor} from '../../components/DocEditor/DocEditor.js';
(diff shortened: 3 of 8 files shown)


stevenle:
i think this feature feels too much like it's solving a use case that i think can be solved in other ways, simpler, more intuitive ways. i also generally would prefer to try to keep the ui as minimal as we can be (so we can avoid microsoft docs style toolbars where there are a million different options).

i think in the project you're working on, a user can currently navigate to a different page and press "c" to edit that page (it opens a new window). 

one alternative option i would consider though is that there currently is an iframe bidirectional channel api where a preview page can signal to the cms certain actions. right now the only action available is "highlight field" which allows a user to click "edit" and it highlights the appropriate field within the cms doc. we can potentially add a new channel api method that says "go to doc Foo/bar" which would give you control over this within the project without having to add new ui elements to the cms. if this route is preferred, feel free to open a new PR!

willthomson:
Thanks — took the channel API route you suggested. Opened #1418.

You were right that the url-guessing was the wrong shape. Letting the page report its own doc id deleted that whole layer: no reverse-mapping from urls, no ambiguity when two collections claim the same path, no existence checks. It also dropped the toolbar toggle — there's no new CMS UI at all now, and a project that doesn't opt in sees no change.

One thing worth flagging separately: `isInPreviewIframe()` only works for the first page the preview pane loads. It checks `document.referrer` against the CMS path, but once someone clicks a link inside the preview the referrer names the previous page of the site, so it returns false from then on. That affects `focusField()` / click-to-edit today, independent of this feature. It's the first commit in #1418, kept separate so it can be reviewed on its own — happy to split it into its own PR if you'd rather land it that way.

PR #1418, opened 2026-09-04
feat: let the preview pane switch the edited doc

willthomson:
Second attempt at #1416, rebuilt around your suggestion instead of the approach you declined. No url guessing, and nothing happens unless a project asks for it.

> [!NOTE]
> Stacked on #1415. Review that first; this targets its branch and I'll rebase onto `main` once it lands.

## What it adds

`PreviewConnection.navigateToDoc(docId)` on the preview channel. A page rendered in the preview pane says which doc it is, and the editor offers to switch to it:

```ts
const preview = new PreviewConnection();
if (preview.isEmbedded) {
  preview.navigateToDoc(doc.id);
}
```

The page already knows its own doc id, so nothing is derived from the url — no reverse-mapping, no ambiguity when two collections claim the same path, no existence probes. That was the weakest part of #1416 and this removes the need for it entirely.

## Confirmation, per your review

The editor asks before switching, since a doc changing on its own is disorienting for someone who doesn't know the page can do that:

> **The preview has changed**
> Do you want to load this page in the editor?
> `Pages/team`
> ☐ Don't ask again this session
> **No**  **Yes**

The checkbox remembers the **answer**, not just the prompt: tick + Yes follows silently for the rest of the session, tick + No stops asking and stops following, so the CMS behaves exactly as it did before this feature. It's scoped to the page session and never persisted — reloading asks again, so there's no stuck state and nothing to undo in settings.

`navigateToDoc(docId, {confirm: false})` skips the prompt, for a site that has its own affordance.

## Nothing happens by default

Two independent conditions, both pre-existing:

- The CMS only acts on the message when `preview.channel` enables the from-preview direction, which defaults to `false`.
- The site has to call `navigateToDoc()` itself.

I deliberately did **not** add a config flag — `preview.channel` already gates the direction and a second key would be redundant. The diff is additive, and `<DocumentPage.Preview>` keeps its `key={docId}`, so existing doc-to-doc navigation is unchanged.

## The doc id is untrusted

It arrives by `postMessage` from whatever page the pane navigated to, so:

- accepted only from this page's own preview iframes (`contentWindow === event.source`, not just an origin check) and from an allowed origin;
- the collection must exist in `__ROOT_CTX.collections`;
- dot segments rejected, route segments `encodeURIComponent`'d.

The slug is deliberately *not* matched against the CMS slug pattern — docs created before it was enforced would silently stop being reachable.

## The first commit is a separate bug fix

`isInPreviewIframe()` tested `document.referrer` against the CMS path, which only holds for the **first** page the pane loads. Follow a link inside the preview and the referrer names the previous page of the site, so it returned `false` for every page after that.

This is pre-existing and affects `focusField()` / click-to-edit today, not just this feature — but the feature can't work without it, which is why it's here rather than in its own PR. Kept as its own commit so it can be reviewed on its own, and happy to split it out if you'd prefer to land it separately. The pane is same-origin, so the parent's location is the reliable signal; the referrer check stays as a fallback.

## Testing

12 unit tests for the message guard, including that it stays distinct from the other un-namespaced messages sharing the channel. 824 pass across `ui`/`shared`/`browser-client`; lint clean.

Exercised against a real site with live Firebase credentials: the prompt and both answers, the session memory in both directions, consecutive follows, back-button behaviour, and the default-off path.

Two things that testing corrected, both worth flagging since I had them wrong in an earlier revision of this description:

- The route is a **push**, not a replace. I had assumed the iframe's own navigation already added a top-level history entry; it doesn't. Replacing discarded the previous doc's entry, so Back appeared to do nothing but flicker the tab title. With push, one Back returns the editor, url, and pane together.
- A follow adds two history entries — the route plus the remounted preview's load. Back still works in a single press, but the second entry is a consequence of keeping `key={docId}`.

## Known tradeoff

Following a link remounts the preview pane, so the page loads a second time. Keeping the pane alive across doc changes fixes that, but it means un-keying `<DocumentPage.Preview>`, which changes behaviour for *all* doc-to-doc navigation. That didn't belong in this PR — happy to propose it separately if you want it.

### DIFF
--- a/.changeset/preview-iframe-detection.md
+++ b/.changeset/preview-iframe-detection.md
@@ -1,5 +1,0 @@
----
-'@blinkk/root-cms': patch
----
-
-fix: detect the preview pane after an in-preview navigation
--- a/.changeset/preview-iframe-detection.md
+++ b/.changeset/preview-iframe-detection.md
@@ -1,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+fix: detect the preview pane after an in-preview navigation
--- a/packages/root-cms/browser-client/browser-client.ts
+++ b/packages/root-cms/browser-client/browser-client.ts
@@ -330,10 +330,26 @@ export class RootCMSBrowserClient {
     return new PreviewConnection();
   }
 
-  /** Returns whether this page is rendered inside the in-CMS preview pane. */
+  /**
+   * Returns whether this page is rendered inside the in-CMS preview pane.
+   *
+   * The parent's location is the reliable signal: the preview iframe is
+   * same-origin, so the CMS path is readable. The referrer only names the CMS
+   * for the first page the pane loads -- once the user follows a link inside
+   * the preview it names the previous page of the site instead -- so it is
+   * kept as a fallback for the cases where the parent can't be read.
+   */
   static isInPreviewIframe(): boolean {
     if (window.parent === window) {
       return false;
+    }
+    try {
+      const parentPath = window.parent.location.pathname;
+      if (parentPath.startsWith('/cms')) {
+        return true;
+      }
+    } catch {
+      // Cross-origin parent: fall through to the referrer check.
     }
     return document.referrer.startsWith(`${window.location.origin}/cms`);
   }
--- a/packages/root-cms/browser-client/browser-client.ts
+++ b/packages/root-cms/browser-client/browser-client.ts
@@ -16,6 +16,7 @@
 
 import {
   isHighlightNodeMessage,
+  NavigateToDocMessage,
   ScrollToDeeplinkMessage,
 } from '../shared/embed-protocol.js';
 import {normalizeSlug} from '../shared/slug.js';
@@ -25,6 +26,7 @@ import {Emitter} from './emitter.js';
 export {SaveState} from '../shared/embed-protocol.js';
 export type {
   HighlightNodeMessage,
+  NavigateToDocMessage,
   RootEmbedMessage,
   ScrollToDeeplinkMessage,
 } from '../shared/embed-protocol.js';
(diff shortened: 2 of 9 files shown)


.changeset/preview-iframe-detection.md
```diff
@@ -0,0 +1,5 @@
+---
+'@blinkk/root-cms': patch
+---
+
+fix: detect the preview pane after an in-preview navigation
```
stevenle: remove this changeset

packages/root-cms/browser-client/browser-client.ts
```diff
@@ -200,6 +202,40 @@ export class PreviewConnection {
...
+   *
+   * Usage:
+   * ```
+   * const preview = new PreviewConnection();
+   * if (preview.isEmbedded) {
+   *   preview.navigateToDoc(doc.id);
+   * }
+   * ```
+   */
+  navigateToDoc(docId: string) {
```
stevenle: now that a developer has programmatic access to navigate to a different doc, the experience can be jarring to a content editor that doesn't know that this is happening, and may be confused why suddenly they're on a different page.

i'm thinking by default, the navigateToDoc() call would open a modal that prompts the user that the page is suggesting the user open another doc, and to confirm the navigation. and maybe there should be an option to this method that bypasses that confirmation message in cases where you have an explicit toggle.

stevenle (commented):
thanks, this pr is looking better than the one previously.

.changeset/preview-iframe-detection.md
willthomson: Done — removed.

packages/root-cms/browser-client/browser-client.ts
willthomson: Added. The editor now confirms before switching, and `navigateToDoc(docId, {confirm: false})` skips it for sites that have their own affordance.

> **The preview has changed**
> Do you want to load this page in the editor?
> `Pages/team`
> ☐ Don't ask again this session
> **No**  **Yes**

One addition beyond what you described: the checkbox remembers the *answer*, not just the prompt. Tick + Yes follows silently for the rest of the session; tick + No stops asking **and** stops following, so the CMS behaves exactly as it did before this feature. It's scoped to the page session and never persisted, so reloading asks again — no stuck state, and nothing needed in settings to undo it.

That came out of testing: with a plain "don't ask again" the checkbox was read only on confirm, so ticking it and clicking No was silently discarded and you'd be prompted again on the next click.

willthomson:
Pushed both changes — replied inline on each.

One correction to flag, since the reasoning was visible in the PR description and it was wrong: the route was doing `replace` on the assumption that the iframe's own navigation had already added a top-level history entry. It hadn't, so replacing discarded the previous doc's entry and the back button appeared to do nothing except flicker the tab title. It's a `push` now — one Back returns the editor, url and preview pane together. Description updated.

Exercised against a real site with live credentials: the prompt and both answers, the session memory in both directions, consecutive follows, and the default-off path.

stevenle (approved):

